PR #24, opened 2015-03-20
Make cameo python 3.4 compatible

phantomas1234:


phantomas1234:
Ok, I am probably going to abandon this attempt in favor of #26

PR #28, opened 2015-04-16
FSEOF and partition

KristianJensen:
I've made an implementation of the FSEOF algorithm. I haven't worked very much with unittests so I didn't really know what to test for apart from checking that the function runs without error...

I also made an alternative partition function that works on all iterables (it still returns a list of lists though, not iterables - I wasn't sure whether that was the point?), and a test for it.

### DIFF
--- a/cameo/core/result.py
+++ b/cameo/core/result.py
@@ -15,15 +15,17 @@
 
 from __future__ import absolute_import, print_function
 
+import getpass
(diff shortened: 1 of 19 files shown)


phantomas1234:
I guess the function should return a little bit more informative output but we can improve this later. Thanks @KristianJensen .

PR #29, opened 2015-04-27
Simulation methods

joaocardoso:
Fix simulation methods.
Increase test units for room according to Shlomi et al 2005.


cameo/util.py
```diff
@@ -32,6 +32,47 @@
...
+        self._verify_model(model)
+        if variable_id in self.variables:
+            if update is not None:
+                update(model, self.variables[variable_id], *args, **kwargs)
+        else:
+            self.variables[variable_id] = create(model, variable_id, *args, **kwargs)
+            model.solver._add_variable(self.variables[variable_id])
+
+    def _verify_model(self, model):
+        assert self._id == hash(model), "Cannot use a different model instance"
```
phantomas1234: asserts are not intended for this (because they can be turned off globaly). @joaocardoso you need add a if statement and raise a proper exception here.

cameo/flux_analysis/simulation.py
```diff
@@ -90,42 +90,40 @@ def pfba(model, objective=None, *args, **kwargs):
...
+           undo=partial(model.solver._remove_constraint, fix_obj_constraint))
+        pfba_obj = model.solver.interface.Objective(add(
+            [mul((sympy.singleton.S.One, variable)) for variable in list(model.solver.variables.values())]),
+            direction='min', sloppy=True)
+        tm(do=partial(setattr, model, 'objective', pfba_obj),
+           undo=partial(setattr, model, 'objective', original_objective))
+        try:
+            solution = model.solve()
+            result = FluxDistributionResult(solution)
+            tm.reset()
```
phantomas1234: @joaocardoso why `tm.reset()` here? That should be covered by the `with` statement.

PR #30, opened 2015-04-28
Better FSEOF output and more tests

KristianJensen:


phantomas1234:
Great! Thanks @KristianJensen.

PR #33, opened 2015-06-18
Fixed bugs where changing bounds could give a value error

KristianJensen:
The following raised errors instead of silently changing other bound:

model.reactions.PGI.lower_bound = 1
model.reactions.PGI.upper_bound = 2
model.reactions.PGI.upper_bound = 0.5
=> ValueError

model.reactions.PGI.upper_bound = -1
model.reactions.PGI.lower_bound = -2
model.reactions.PGI.upper_bound = -0.5
=> ValueError


phantomas1234:
Awesome, thank you @KristianJensen!

PR #34, opened 2015-07-07
Use appveyor to build windows wheels for cameo

phantomas1234:


phantomas1234:
I realized that cameo doesn't need a windows wheel because it doesn't include any C extensions. Instead swiglpk needs appveyor integration to generate a windows wheel.

PR #63, opened 2016-09-12
Feature/phenopp yield

hredestig:


cameo/flux_analysis/analysis.py
```diff
@@ -367,6 +381,101 @@ class _PhenotypicPhasePlaneChunkEvaluator(object):
...
+                            reaction in self.model.medium.reaction_id]
+        source_reactions = [reaction for reaction in medium_reactions if
+                            (self.n_carbon_reaction(reaction) > 0) and
+                            (reaction.flux < 0)]
+        if len(source_reactions) == 0:
+            raise NoSourceError()
+        return source_reactions
+
+    @classmethod
+    def n_carbon(cls, metabolite):
```
sgalkina: should probably be the method of Metabolite class

cameo/core/metabolite.py
```diff
@@ -46,6 +46,13 @@ def remove_from_model(self, method="subtractive", **kwargs):
         super(Metabolite, self).remove_from_model(method, **kwargs)
         model.solver.remove(model.solver.constraints[self.id])
 
+    def n_carbon(self):
+        """ :return: int, number of carbons in metabolite"""
+        try:
+            return self.elements['C']
+        except KeyError:
+            return 0
```
sgalkina:
```python
return self.elements.get('C', 0)
```

cameo/core/metabolite.py
```diff
@@ -47,6 +47,15 @@ def remove_from_model(self, method="subtractive", **kwargs):
         super(Metabolite, self).remove_from_model(method, **kwargs)
         model.solver.remove(model.solver.constraints[self.id])
 
+    def n_carbon(self):
```
joaocardoso: Shouldn't this be a property of the metabolite?

cameo/core/reaction.py
```diff
@@ -492,6 +492,16 @@ def change_bounds(self, lb=None, ub=None, time_machine=None):
         if ub is not None:
             self.upper_bound = ub
 
+    def n_carbon(self):
```
joaocardoso: Same as before.

cameo/flux_analysis/analysis.py
```diff
@@ -367,6 +378,100 @@ class _PhenotypicPhasePlaneChunkEvaluator(object):
...
+        Returns
+        -------
+        float
+            the mol carbon atoms in the product (as defined by the model objective) divided by the mol carbon in the
+            input reactions (as defined by the model medium) """
+        try:
+            carbon_input_flux = self.total_flux(self.c_source_reactions)
+            carbon_output_flux = self.total_flux([self.product_reaction])
+            return carbon_output_flux / (carbon_input_flux * -1)
+        except (Infeasible, UndefinedSolution, ZeroDivisionError):
```
joaocardoso: UndefinedSolution is an error that should be raised. There is something going on with GLKP not returning a proper error if presolve is off, but UndefinedSolution should break. We should only trap infeasible solutions. The other types should be raise because they result from bad model/model constraints.

cameo/flux_analysis/analysis.py
```diff
@@ -367,6 +378,100 @@ class _PhenotypicPhasePlaneChunkEvaluator(object):
...
+                return 0
+            source_flux = self.total_flux(self.c_source_reactions, False)
+            product_flux = self.total_flux([self.product_reaction], False)
+            mol_prod_mol_src = product_flux / (source_flux * -1)
+            source_metabolite = list(self.c_source_reactions[0].metabolites)[0]
+            product_metabolite = list(self.product_reaction.metabolites)[0]
+            product_mass = product_metabolite.formula_weight
+            source_mass = source_metabolite.formula_weight
+            return (mol_prod_mol_src * product_mass) / source_mass
+        except (Infeasible, UndefinedSolution, ZeroDivisionError):
```
joaocardoso: Same as before with the UndefinedSolution

cameo/flux_analysis/analysis.py
```diff
@@ -468,9 +580,47 @@ def data_frame(self):
...
+                              'c_yield': ('c_yield_upper_bound',
+                                          'c_yield_lower_bound',
+                                          'mol C product / mol C source')}
+        if estimate not in possible_estimates:
+            raise Exception('estimate must be one of %s' %
+                            ', '.join(possible_estimates.keys()))
+        upper, lower, description = possible_estimates[estimate]
         if title is None:
-            title = "Phenotypic Phase Plane"
+            title = "Phenotypic Phase Plane ({})".format(description)
```
joaocardoso: The same way you change the title, I think it will be nice to also append the units to the x and y axis labels. The assumption before was that we were working only with fluxes.

cameo/flux_analysis/analysis.py
KristianJensen: The GLPK UndefinedSolution issue was fixed in optlang a while ago. Is it still a problem?
Also why is Infeasible excepted? As far as I can see the try block doesn't attempt to solve the model. And even if the model is infeasible, wouldn't the yield be undefined rather than just 0?

cameo/flux_analysis/analysis.py
```diff
@@ -367,6 +378,100 @@ class _PhenotypicPhasePlaneChunkEvaluator(object):
...
+            summed flux for a set of reactions. Possibly the flux of carbon """
+        flux = 0
+        for rxn in reactions:
+            carbon = 1
+            if by_carbon:
+                carbon = sum(metabolite.n_carbon() for metabolite in rxn.reactants)
+            try:
+                flux += rxn.flux * carbon
+            except AssertionError:
+                # optimized flux can be out of bounds, in that case treat as zero flux
```
KristianJensen: The assertion error is also sometimes raised even though the flux is not zero, e.g. flux = -10.001 when lb=-10. In that case this method would erroneously return a total flux of 0.

cameo/flux_analysis/analysis.py
hredestig: true, not needed / appropriate here, fixed

cameo/flux_analysis/analysis.py
```diff
@@ -367,6 +379,100 @@ class _PhenotypicPhasePlaneChunkEvaluator(object):
...
+
+        Returns
+        -------
+        list
+           list of media reaction currently with negative flux or 0 if non defined
+        """
+        medium_reactions = [self.model.reactions.get_by_id(reaction) for reaction in self.model.medium.reaction_id]
+        self.model.solve()
+        source_reactions = [reaction for reaction in medium_reactions if
+                            (reaction.n_carbon > 0) and (reaction.flux < 0)]
```
joaocardoso: For many reasons concerning safe usage you should not use *reaction.flux*.
The solution object generated <code>_production_envelope_inner</code> should be an argument of all of this methods you implemented.
The *solution* object contains the *.fluxes* attribute that will contain the fluxes calculated. By doing that you don't need to concern about catching the Exceptions raised during optimizing the model.

The code in  <code>_production_envelope_inner</code> becomes something like this.

<code>

    self.model.objective.direction = 'min'

    try:

        solution = self.model.solve().f

    try:

        solution = self.model.solve()
        interval.append(solution.f)

        interval_carbon_yield.append(self.carbon_yield(solution))

        interval_mass_yield.append(self.mass_yield(solution))

    except (Infeasible, UndefinedSolution):

        interval.append(0)

        interval_carbon_yield.append(0)

        interval_mass_yield.append(0)
    self.model.objective.direction = 'max'
    try:

        solution = self.model.solve()

        interval.append(solution.f)

        interval_carbon_yield.append(self.carbon_yield(solution))

        interval_mass_yield.append(self.mass_yield(solution))

    except (Infeasible, UndefinedSolution):

        interval.append(0)

        interval_carbon_yield.append(0)

        interval_mass_yield.append(0)

</code>

phantomas1234: Keep in mind that using `solution.fluxes` will be a lot slower than `reaction.flux` because it will retrieve all fluxes. So I might be in favor of the existing code.

joaocardoso: True. My concern is parallelization because reaction.flux will fetch directly from the state of the solver. Ok, I agree on keeping *reaction.flux* until we figure out a better implementation. But there is no need to except any kind of SolverError inside *carbon_yield* and *mass_yield*.

hredestig: cheers @joaocardoso, learnt something, looks much better to me, however sadly yes it does get substantially slower that way..

using `solution.fluxes`
```{python}
In[112]: %time ppp2d = phenotypic_phase_plane(model, ['EX_o2_LPAREN_e_RPAREN_', 'EX_co2_LPAREN_e_RPAREN_'])

CPU times: user 17 s, sys: 80.3 ms, total: 17.1 s
Wall time: 17.2 s
In[113]: %time ppp = phenotypic_phase_plane(model, ['EX_o2_LPAREN_e_RPAREN_'])

CPU times: user 3.66 s, sys: 25 ms, total: 3.69 s
Wall time: 3.74 s
```

using `reaction.flux`
```{python}
In[114]: %time ppp = phenotypic_phase_plane(model, ['EX_o2_LPAREN_e_RPAREN_'])

CPU times: user 72.4 ms, sys: 1.73 ms, total: 74.2 ms
Wall time: 74.8 ms
In[115]: %time ppp2d = phenotypic_phase_plane(model, ['EX_o2_LPAREN_e_RPAREN_', 'EX_co2_LPAREN_e_RPAREN_'])

CPU times: user 867 ms, sys: 11 ms, total: 878 ms
Wall time: 881 ms
```

joaocardoso: And I edited my first comment to remove the UndefinedSolution. If UndefinedSolution exceptions start to pop up, something else is wrong...

KristianJensen: Maybe the relevant fluxes (source and product reactions) could be fetched and stored immediately after the model is solved during __init__. That way all the unnecessary fluxes won't be retrieved and there is no risk of the model changing between solving it and getting the fluxes.

cameo/flux_analysis/analysis.py
```diff
@@ -367,6 +378,100 @@ class _PhenotypicPhasePlaneChunkEvaluator(object):
...
+        Returns
+        -------
+        float
+            the mol carbon atoms in the product (as defined by the model objective) divided by the mol carbon in the
+            input reactions (as defined by the model medium) """
+        try:
+            carbon_input_flux = self.total_flux(self.c_source_reactions)
+            carbon_output_flux = self.total_flux([self.product_reaction])
+            return carbon_output_flux / (carbon_input_flux * -1)
+        except (Infeasible, ZeroDivisionError):
```
KristianJensen: Am I completely missing something or is the model not solved in this try block? Otherwise except Infeasible is not necessary.

cameo/flux_analysis/analysis.py
phantomas1234: @joaocardoso I don't think `reaction.flux` is an issue in parallelization because models are pickled and unpickled, right? So it is "thread safe"?

cameo/flux_analysis/analysis.py
```diff
@@ -376,20 +479,27 @@ def _production_envelope_inner(self, point):
             for (reaction, coordinate) in zip(self.variable_reactions, point):
                 reaction.change_bounds(coordinate, coordinate, time_machine=tm)
             interval = []
+            interval_carbon_yield = []
+            interval_mass_yield = []
             tm(do=int, undo=partial(setattr, self.model.objective, 'direction', self.model.objective.direction))
             self.model.objective.direction = 'min'
             try:
                 solution = self.model.solve().f
             except (Infeasible, UndefinedSolution):  # Hack to handle GLPK bug
```
KristianJensen: The GLPK bug has been fixed, so UndefinedSolution shouldn't be excepted

cameo/flux_analysis/analysis.py
```diff
@@ -367,6 +378,98 @@ class _PhenotypicPhasePlaneChunkEvaluator(object):
...
+        only defined when we have only one product (as defined by the model
+        objective) and only one compound as carbon source (as defined by the
+        model medium).
+
+        Returns
+        -------
+        float
+            gram product per 1 g of feeding source or 0 if more than one product or feeding source
+        """
+        try:
```
hredestig: fixed

cameo/flux_analysis/analysis.py
hredestig: fixed

cameo/flux_analysis/analysis.py
```diff
@@ -367,29 +378,139 @@ class _PhenotypicPhasePlaneChunkEvaluator(object):
...
         self.model = model
         self.variable_reactions = variable_reactions
+        objective_reactions = [reaction for reaction in model.reactions
+                               if reaction.objective_coefficient != 0]
+        if len(objective_reactions) != 1:
+            raise Exception('multiple objectives not supported')
+        self.product_reaction = objective_reactions[0]
+        self.c_source_reactions = self.get_c_source_reactions()
+
+    def get_c_source_reactions(self):
```
joaocardoso: Hi guys. This is a tricky and dangerous one.

You are changing the objective of the problem during FVA. When you have two carbon sources, but one cannot be converted into biomass, but it is a precursor of something else, you don't know before hand if something is maximizing just a specific set reactions just by maximizing biomass. Because the total flux is a sum of (n_carbon * flux), you just need to take the *model.medium* reactions *if r.n_carbon > 0*. The overhead is small and avoids problems with evaluating "exotic" designs.

@phantomas1234 @KristianJensen @meono: any thoughts?

joaocardoso: Ok.

This network:

-> A -> B -> C -> Biomass
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;|
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;D
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;| - (irreversible from D->F)
&nbsp;&nbsp;-> F->E -> Product

If you maximize biomass, F cannot go to biomass, therefore F.flux is zero when you maximize biomass. When you maximize product F is also a carbon source for the process, however is not taken into account with the initial filter.

phantomas1234: The objective is not changed during FVA/phenotypic_phase_plane. It is always what `objective=` specifies. Also, if you have multiple carbon sources you cannot compute mass or molar yields, only c-mol/c-mol yields. In your example it would simply be `product_flux/(A_flux * A.n_carbons + F_flux * F.n_carbons)`.

hredestig:
the first problem with variables as reactions are sorted now, but I'm unsure what should be the right behaviour for mass/carbon yield should be when the medium reactions with negative flux includes or doesn't include e.g. co2 depending on how the rest of the fluxes are.. before my last commit: only solve the model once and get the c source reactions at that point. Then, in your example, co2 was included leading to NA mass yield (only defined for single c source). Now, I update the c-source after the bounds have been set according to the point envelope. Then, we sometimes get glc, sometimes glc + co2. Seems like current behaviour fits with the model but it isn't intuitive...

phantomas1234:
Thanks @hredestig, looks good.

PR #64, opened 2016-09-19
feat: adapt model to real experiments, helper class

sgalkina:
Functionality of creating and annotating new metabolites and their exchange/demand reactions is required by multiple services. Considering that it depends only on ```cameo``` and operates on cameo models, it makes a good fit as one of the higher-level api modules.
@hredestig @phantomas1234

### DIFF
--- a/cameo/api/adapter.py
+++ b/cameo/api/adapter.py
@@ -1,0 +1,150 @@
+import re
+from cameo.data import metanetx
+from cameo.core.reaction import Reaction
+from cameo.core.metabolite import Metabolite
+import logging
+logger = logging.getLogger(__name__)
+
+
+def clean_bigg_id(string):
+    return re.sub(r"bigg:|dsh", "", string)
+
+
+def get_existing_metabolite(mnx_id, model, compartment):
+    """Find compartment in the model by Metanetx id.
+
+    Parameters
+    ----------
+    mnx_id - Metanetx id
+    model - cobra model
+    compartment - f.e "_c"
+
+    Returns
+    -------
+    Metabolite or None
+
+    """
+    if not mnx_id:
+        return
+    try:
+        clean_id = clean_bigg_id(metanetx.mnx2bigg[mnx_id])
+        return model.metabolites.get_by_id(clean_id + compartment)
+    except KeyError:
+        try:
+            return model.metabolites.get_by_id(mnx_id + compartment)
+        except KeyError:
+            pass
+
+
+def contains_carbon(metabolite):  # TODO: use method from Metabolite class when this change is merged
+    if not metabolite.formula:
+        raise ValueError("No formula for metabolite {}, it's unknown if there is carbon in it")
+    return 'C' in metabolite.elements
+
+
+def find_metabolite_info(met_id):
+    """Find chemical formula of metabolite in metanetx.chem_prop dictionary
+
+    :param met_id: string of format "<metabolite_id>_<compartment_id>", where <metabolite_id> is a BIGG id or a
+    Metanetx id
+
+    :return:
+    """
+    met_id = met_id[:-2]
+    try:
+        if met_id in metanetx.chem_prop.index:
+            return metanetx.chem_prop.loc[met_id]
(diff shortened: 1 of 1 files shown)


cameo/api/adapter.py
```diff
@@ -0,0 +1,150 @@
...
+def contains_carbon(metabolite):  # TODO: use method from Metabolite class when this change is merged
+    if not metabolite.formula:
+        raise ValueError("No formula for metabolite {}, it's unknown if there is carbon in it")
+    return 'C' in metabolite.elements
+
+
+def find_metabolite_info(met_id):
+    """Find chemical formula of metabolite in metanetx.chem_prop dictionary
+
+    :param met_id: string of format "<metabolite_id>_<compartment_id>", where <metabolite_id> is a BIGG id or a
```
hredestig: i guess better use google (?) style docstrings like rest of cameo (i just updated a pr to adhere to that myself (:)

sgalkina:
> what's the scope of 'api'? perhaps it fits better in 'models'?

@hredestig currently `models` module only contains models source files - `.xml` and `.json`, and code to reach them. So I'm not sure it would be right place to keep model modification class.

Another possible place could be `solver_based_model` in `core`, but functions introduced in this pr are not of a basic functionality, and more of high-level encapsulations using those basics functions from `core` module.

cameo/api/adapter.py
```diff
@@ -0,0 +1,181 @@
...
+    met_id = met_id[:-2]
+    try:
+        if met_id in metanetx.chem_prop.index:
+            return metanetx.chem_prop.loc[met_id]
+        return metanetx.chem_prop.loc[metanetx.all2mnx['bigg:' + met_id]]
+    except KeyError:
+        return None
+
+
+class ModelModification(object):
```
sgalkina: @phantomas1234 @hredestig now this class does not use TimeMachine anywhere - would it make sense to use it, initializing time machine in constructor? is there any best practise recommendation of using time machine for all the model modification functionality?

hredestig: I see these new models that `ModelModification` creates as really new model objects to do further analysis on. These further analysis will likely use `TimeMachine` to avoid further copying, however using `TimeMachine` also in the constructor here appears excessive and confusing. I suggest let's implement that if the need ever appears.

More important I guess would be a couple of small unit tests if you care to add that?

hredestig:
this looks great to me now - is it ready to merge?

cameo/api/adapter.py
```diff
@@ -0,0 +1,181 @@
...
+
+        """
+        try:
+            logger.debug('Add exchange reaction for metabolite: {}'.format(metabolite.id))
+            exchange_reaction = self.model.add_exchange(metabolite, prefix='EX_')
+            self.added_reactions.add(exchange_reaction.id)
+        except ValueError:
+            logger.debug('Exchange reaction exists for metabolite {}'.format(metabolite.id))
+
+    def add_adapter_reaction(self, metabolite, existing_metabolite, gene_name):
```
joaocardoso: Why adapter reactions need genes?

joaocardoso:
I think this functionality is out of **cameo** scope, we should consider migrating this and other stuff to cobrapy. This applies to _essential_reactions_ and _essential_genes_.

cameo/api/adapter.py
sgalkina: To annotate the reaction with gene name
```python
adapter_reaction.gene_reaction_rule = gene_name
```

joaocardoso: That I understood, but adapter reactions are just virtual reactions that connect a metabolite from a model to another model/database so that they are the same right? That reaction is not a representation of an existing reaction carried by a protein so it should not be annotated with a gene reaction rule.

sgalkina: great point, thanks

hredestig:
build should work now if you merge with devel

sgalkina:
Moved to https://github.com/biosustain/driven

Issue #65, opened 2016-09-19
implement optswap


hredestig:
optswap [1] is an algorithm for recommending co-factor swaps for strain design - let's implement this in cameo using the pattern also used for knockout optimisation

[1] King ZA, Feist AM. Optimizing Cofactor Specificity of Oxidoreductase Enzymes
     for the Generation of Microbial Production Strains—OptSwap. Ind
     Biotechnol. 2013;9:
     236–246. doi:[10.1089/ind.2013.0005](https://doi.org/10.1089/ind.2013.0005)

PR #67, opened 2016-09-22
Feature/optswap

hredestig:
Optimize co-factor swapping #65 

flux through a given reaction can sometimes be optimized by swapping complementary
co-factor. This class implements a search for reactions when swapped improve the given objective. Briefly,
the approach is to

 - find reactions that can safely be swapped e.g. (nad_c -> nadp_c, nadh_c -> nadph_c) without resulting in an
   infeasible model

- add reactions that have the co-factors swapped and then by a search algorithm switching one off in favor of the
  other


cameo/strain_design/heuristic/evolutionary/optimization.py
```diff
@@ -358,12 +362,42 @@ def _calculate_fitness(self, solution, decoded):
             return self.objective_function(self.model, solution, decoded)
 
 
+class SwapEvaluator(KnockoutEvaluator):
```
joaocardoso: I think we can have a generic Evaluator. I will be conceptually easier and later on we can extend evaluators that are not Knockout related.

cameo/strain_design/heuristic/evolutionary/optimization.py
```diff
@@ -806,3 +845,223 @@ def __init__(self, reaction_db=None, *args, **kwargs):
...
+       a list of length 2 where the first element contains a list of metabolite identifiers to swap-out by replacing
+       them with the metabolites listed in the second element of the same list, see e.g. `NADH_NADPH` which is also
+       the default
+    candidate_reactions : list
+       reactions to consider for co-factor swap - if not given then search for all reactions that include the given
+       cofactors
+    skip_reactions : list
+       reactions to not consider for co-factor swap, defaults to the objective function if not provided
+    """
+    def __init__(self, model, cofactor_id_swaps, candidate_reactions=None, skip_reactions=None):
```
joaocardoso: Shouldn't we automatically figure out the candidate reactions automatically based on the swap pattern?

cameo/strain_design/heuristic/evolutionary/optimization.py
```diff
@@ -806,3 +845,223 @@ def __init__(self, reaction_db=None, *args, **kwargs):
...
+
+        if time_machine is not None:
+            time_machine(do=partial(do_change, from_reaction, to_reaction,
+                                    from_reaction.upper_bound, from_reaction.lower_bound),
+                         undo=partial(do_change, to_reaction, from_reaction,
+                                      from_reaction.upper_bound, from_reaction.lower_bound))
+        else:
+            do_change(from_reaction, to_reaction, from_reaction.upper_bound, from_reaction.lower_bound)
+
+    def find_swappable_reactions(self, threshold=1e-6):
```
joaocardoso: You don't need to worry about this. A swap on it's on might be infeasible but in combination with other swaps it might work. The algorithm will discard those solutions on its own. This leads to a decrease of possible solutions.

cameo/strain_design/heuristic/evolutionary/optimization.py
```diff
@@ -806,3 +845,223 @@ def __init__(self, reaction_db=None, *args, **kwargs):
...
+    def __init__(self, model, cofactor_id_swaps=NADH_NADPH, candidate_reactions=None,
+                 skip_reactions=None, *args, **kwargs):
+        super(self.__class__, self).__init__(model=model, knockout_evaluator=SwapEvaluator, *args, **kwargs)
+        self.model = SwapperModel(model, cofactor_id_swaps, candidate_reactions, skip_reactions)
+        self._ko_type = SWAP_KNOCKOUT_TYPE
+        self.representation = list(self.model.swapped_reactions.keys())
+        self._decoder = decoders.ReactionKnockoutDecoder(self.representation, self.model)
+
+
+class SwapperModel(SolverBasedModel):
```
joaocardoso: I am not sure about extending the model. The method that Reaction.add_metabolites should be efficient. I think that you could do this with a single swap function.

cameo/strain_design/heuristic/evolutionary/optimization.py
hredestig: I'll leave that for now as it required refactoring previous methods as well.

cameo/strain_design/heuristic/evolutionary/optimization.py
hredestig: I understand it might seem unnecessary and there are for sure many ways to do this - however, swapping a reaction as a method of a model seemed oop and convenient to me as it manages all the arguments in a straight-forward fashion, avoiding reliance on `*args, **kwargs` and ilk - I did not notice significant draw-backs with subclassing the model - did you?

cameo/strain_design/heuristic/evolutionary/optimization.py
hredestig: it does actually :) see call to `find_swappable_reactions` in `__init__`

cameo/strain_design/heuristic/evolutionary/optimization.py
hredestig: My goal was to reduce the search space as it already can get quite enormous thus many solutions will never be tested - but fair enough, this step is not well motivated, I'll remove this check

joaocardoso: I understood that. It makes sense with knockouts, because they have a cumulative effect. However, combinations of swaps might (say you change two reactions that are dependent via their cofactors) will work.

cameo/strain_design/heuristic/evolutionary/optimization.py
joaocardoso: True! My bad, sorry. At 22h is hard to follow all the details.

cameo/strain_design/heuristic/evolutionary/optimization.py
hredestig: gotcha - fixed now

cameo/strain_design/heuristic/evolutionary/optimization.py
```diff
@@ -377,7 +411,9 @@ def __init__(self, simulation_method=pfba, wt_reference=None, *args, **kwargs):
         self._simulation_method = None
         self.simulation_method = simulation_method
         self.representation = None
-        self._knockout_evaluator = None
+        if knockout_evaluator is None:
+            knockout_evaluator = KnockoutEvaluator
+        self.knockout_evaluator = knockout_evaluator
```
sgalkina:
```python
self.knockout_evaluator = knockout_evaluator or KnockoutEvaluator
```

cameo/strain_design/heuristic/evolutionary/optimization.py
joaocardoso: IDK. I'll need to think about it. My current concerns are:
1. whether to have a copy of the reaction and swap bounds like you do or to use the Reaction.add_metabolites.
2. mainting another extension of the Model class. If later on I would like to swap a cofactor on my own (outside of this optimization) I would like to have a method like _swap_cofactors(reaction_id, cofactors, time_machine)_ that is independent of this whole optswap process.

cameo/strain_design/heuristic/evolutionary/optimization.py
hredestig: ah right, that's prettier, cheers

cameo/strain_design/heuristic/evolutionary/optimization.py
```diff
@@ -806,3 +845,209 @@ def __init__(self, reaction_db=None, *args, **kwargs):
...
+
+class SwapperModel(SolverBasedModel):
+    """An extension of the `SolverBasedModel` to add functions to easily add reactions that have co-factors swapped
+    etc.
+
+    Attributes
+    ----------
+    model : SolverBasedModel
+       the model to convert
+    cofactor_id_swaps : list
```
sgalkina: If it is certain that `cofactor_id_swaps` list is always going to be of the same length, tuple as a data sctructure makes a better fit. 
Talking about tuples, looks like  `cofactor_id_swaps` has strict definition of what is the first and what is the second element. In this case, `namedtuple` from `collections` module fits better
```python
>>> from collections import namedtuple
>>> Swaps = namedtuple('Swaps', ['swap_out', 'swap_in'])
>>> cofactor_id_swaps = Swaps(['nad_c', 'nadh_c'], ['nadp_c', 'nadph_c'])
>>> cofactor_id_swaps.swap_out
['nad_c', 'nadh_c']
>>> cofactor_id_swaps.swap_in
['nadp_c', 'nadph_c']
```
Which more explicit than seeing `cofactor_id_swaps[0]` in the code, like in `swap_search` function.

hredestig: true, changing to tuple, however since the co-factor pair is interchangeable we don't think of swap_in and swap_out just two pairs that can be changed for each other hence I stick to the unnamed tuple

cameo/strain_design/heuristic/evolutionary/optimization.py
```diff
@@ -806,3 +845,209 @@ def __init__(self, reaction_db=None, *args, **kwargs):
...
+            the identifier of the reaction to swap
+        time_machine : TimeMachine
+            enable undo of the swap
+        """
+        from_reaction = self.reactions.get_by_id(reaction_id)
+        to_reaction = self.swapped_reactions[reaction_id]
+        if abs(from_reaction.lower_bound) < 1e-6 and abs(from_reaction.upper_bound) < 1e-6:
+            tmp_reaction = from_reaction
+            from_reaction = to_reaction
+            to_reaction = tmp_reaction
```
sgalkina:
```python
from_reaction, to_reaction = to_reaction, from_reaction
```

cameo/strain_design/heuristic/evolutionary/optimization.py
```diff
@@ -43,7 +45,9 @@
 __all__ = ['ReactionKnockoutOptimization', 'GeneKnockoutOptimization']
```
phantomas1234: add `CofactorSwapOptimization` to `__all__`. Also change `from cameo.strain_design.heuristic.evolutionary.optimization import ReactionKnockoutOptimization, \
    GeneKnockoutOptimization` in evolutionary/__init__.py to `from cameo.strain_design.heuristic.evolutionary.optimization import *`

cameo/strain_design/heuristic/evolutionary/optimization.py
```diff
@@ -806,3 +843,205 @@ def __init__(self, reaction_db=None, *args, **kwargs):
 
     def run(self, **kwargs):
         KnockinKnockoutOptimization.run(self, **kwargs)
+
+
+class CofactorSwapOptimization(KnockoutOptimization):
```
phantomas1234: Need to add unit tests for `CofactorSwapOptimization`.

hredestig: column name results fixed, and plot raises `NotImplementedError`. I can't reproduce the problem with the non-appearing plot, however that and the missing `plot` function is older code that needs fixing so can work on that in a bug fix issue (e.g. not even sure we need that plot function so would rather drop it but that depends on what the original intent was).

PR #69, opened 2016-09-26
Fix: gene.knock_out method

joaocardoso:
The method was not implemented to support OR relationships between genes and reactions.


joaocardoso:
Unit tests are still missing, I will add them soon

cameo/core/reaction.py
```diff
@@ -187,6 +188,13 @@ def __deepcopy__(self, memo):
     def lower_bound(self):
         return self._lower_bound
 
+    @property
+    def functional(self):
```
hredestig: add a docstring here, although the name is quite descriptive a definition would help, e.g, a gene is functional unless the gpr states otherwise

cameo/core/gene.py
```diff
@@ -74,13 +74,20 @@ def knock_out(self, time_machine=None):
...
-                reaction.knock_out()
+        if time_machine:
+            time_machine(do=partial(setattr, self, 'functional', False),
+                         undo=partial(setattr, self, 'functional', True))
+        else:
+            self.functional = False
+
+        for reaction in self.reactions:
+            if not reaction.functional:
+                def _(reaction, lb, ub):
```
sgalkina: can `reaction.change_bounds` be used in this case?

phantomas1234: yes, this should use `change_bounds`.

PR #70, opened 2016-09-30
feat: essential metabolites

joaocardoso:
This should go to cobrapy when refactored like I mentioned in #64 .


phantomas1234:
@joaocardoso Can you please fix the unit test coverage issue.

cameo/core/metabolite.py
```diff
@@ -53,6 +54,45 @@ def constraint(self):
...
+        ----------
+        .. [1] Kim, P.-J., Lee, D.-Y., Kim, T. Y., Lee, K. H., Jeong, H., Lee, S. Y., & Park, S. (2007).
+          Metabolite essentiality elucidates robustness of Escherichia coli metabolism. PNAS, 104(34), 13638-13642
+        """
+        # restrict reactions to produce metabolite
+        for reaction in self.reactions:
+            if reaction.metabolites[self] > 0:  # for positive stoichiometric coefficient set lb to 0
+                reaction.change_bounds(lb=0, time_machine=time_machine)
+            elif reaction.metabolites[self] < 0:  # for negative stoichiometric coefficient set ub to 0
+                reaction.change_bounds(ub=0, time_machine=time_machine)
```
hredestig: @joaocardoso  if reaction has original bounds e.g. (100, 50) then we do this here, we get (0, 50) which doesn't make sense and cameo's upper_bound setter is friendly to convert to (0, 0) [but disregarding the time machine].  what should be the intended bound in this case? (0, 0)?

joaocardoso: Yes, if the lower bound is also positive them it should be (0, 0). I will add a fix. Same applies if it is in the substrates. If the bounds are (-100, -50), it should be 0, 0. If the bounds are (-100, 50) it should become (0, 50). Working on it.

KristianJensen: In that case there would be two inconsistent constraints on the flux through the reaction (at least 50 from the bounds and non-positive from knocking out the metabolite). Wouldn't it make more sense if the model was then infeasible? Otherwise this method will silently overrule constraints imposed at an earlier point

joaocardoso: @KristianJensen: what would be the equivalent of knockout a reaction that has certain bounds? The bounds would become (0, 0) right? So the point is that you cannot run any reaction that consumes this metabolite. If there is a constraint on reactions producing it, it will remain unchanged.

hredestig: phrase it like that in the doc, much clearer than the snippet from the paper..

KristianJensen: I would still argue that the new constraint (no flux through reactions that consume the metabolite) should be added on top of existing constraints (at least 50 flux through reaction) instead of overriding it silently. The same should be the case for knocking out a reaction imo, even though I know that's not how it currently behaves.

joaocardoso: Ok, I understand what you say. We made sense of that before. The problem atm is to fix priorities. The constraint can make the solution infeasible really easy: if the bounds are (50, 100) for R1 and the constraint 0 <= R1 <= 0. This makes the solution infeasible and that is not the same as no growth. It also doesn't reflect the effect of a knockout. This is tricky. I would go with implement this way now and later figure out this priorities. @KristianJensen @hredestig  do we agree?

hredestig: I agree, this behavior is for better or worse consistent with knocking genes and so changing this to instead have additional constraints rather than just adjusting the bounds can be contributed as a new feature

PR #74, opened 2016-10-10
Bugfix/test flux

hredestig:
bugfix: enable testing for fva/ppp and debug 2dppp 

- unit test based on data frames in test_flux_analysis were broken due
  to mistake in assert function which could not throw a AssertionError
  as it should - now this is fixed
- multi-dimensional phenotypic phase plane was broken in that the order
  of the coordinates did not match the the labels of the reactions, this
  has been fixed
- new reference data sets has been created


tests/data/REFERENCE_PPP_o2_EcoliCore_ac.csv
```diff
@@ -0,0 +1,21 @@
+,EX_o2_LPAREN_e_RPAREN_,objective_lower_bound,objective_upper_bound
```
hredestig: @KristianJensen this is now used as reference in `test_one_variable_sequential_metabolite` which previously used `REFERENCE_PPP_o2_EcoliCore`. Since this test changes the objective to a demand reaction for `o2_e` it did yield result identical to `REFERENCE_PPP_o2_EcoliCore` which has biomass as objective. I changed this to ac_e and made a new reference, could you please confirm/suggest changes

tests/data/REFERENCE_PPP_o2_glc_EcoliCore.csv
```diff
@@ -1,401 +1,401 @@
 ,EX_o2_LPAREN_e_RPAREN_,EX_glc_LPAREN_e_RPAREN_,objective_lower_bound,objective_upper_bound
-0,-60.00000000000003,-10.0,0.0,-1.0867823178038478e-15
```
hredestig: @phantomas1234 the reference 2d-ppp has listed objective upper bound as negative (optimizing for biomass) which was a mistake? or am I misunderstanding something? I could not reproduce this result so have updated with new reference, please confirm (or suggest correction)

phantomas1234:
@hredestig Why does codecov/patch not pass? Should I merge this anyways?

hredestig:
I don't know. Looks like noise from codecov when inspecting the details - like misinterpretation of lines / versus statements.. Yes, new ppp2d result makes sense to you (and the old one looks like a mistake) then we can merge this.

PR #75, opened 2016-10-12
Fix bugs when changing bounds

KristianJensen:
I found a couple of bound bugs.

Particularly change_bounds when used with TimeMachine would not correctly reset a bound that was implicitly changed.


KristianJensen:
@hredestig I assigned you on this one. It's just fixes for a couple of small bugs and some tests that would have caught these bugs. Shouldn't take long to review.

tests/test_solver_based_model.py
```diff
@@ -252,6 +252,22 @@ def test_set_bounds_scenario_3(self):
             self.assertEqual(reac.lower_bound, -11)
             self.assertEqual(reac.upper_bound, 2)
 
+        def test_set_bounds_scenario_4(self):
+            reac = self.model.reactions.ACALD
+            reac.knock_out()
+            reac.lower_bound = 2
+            self.assertEqual(reac.lower_bound, 2)
+            self.assertEqual(reac.upper_bound, 2)
```
hredestig: this upper_bound change happens implicitly, which is good and all, could we just have the unit-tests a bit more explicit setting the bounds first, e.g. `reac._lower_bound = 0; reac._upper_bound = 1` first as this would also make us safe to changes in the test model later on

KristianJensen: You mean instead of doing reac.knock_out() ?

PR #76, opened 2016-10-13
Feature - improved ProblemCache

joaocardoso:
New improvments are:
1. Use _with_ statement.
2. Use new optlang API (_add_ and _remove_) instead of protected methods
3. More documentation


tests/test_util.py
```diff
@@ -54,6 +61,43 @@ def test_with_statement(self):
...
+        lmoma(model=TESTMODEL, reference=self.reference)
+
+        # After running lmoma without cache, the number of variables and constraints remains the same
+        self.assertEqual(self.n_constraints, len(TESTMODEL.solver.constraints))
+        self.assertEqual(self.n_variables, len(TESTMODEL.solver.variables))
+
+        cache = ProblemCache(TESTMODEL)
+        lmoma(model=TESTMODEL, reference=self.reference, cache=cache)
+        # After running lmoma with cache, the number of variables is 2 times the original number
+        self.assertEqual(2*self.n_variables, len(TESTMODEL.solver.variables))
```
KristianJensen: Can this be tested in a way that's not dependent on the exact implementation of lmoma?
Maybe with a toy method? Otherwise at least the test should be renamed to include "lmoma", so when this method fails at one point it's clear that it could have something to do with lmoma.

tests/test_util.py
```diff
@@ -54,6 +61,43 @@ def test_with_statement(self):
...
+        cache.reset()
+        # After reset cache, the problem should return to its original size
+        self.assertEqual(self.n_constraints, len(TESTMODEL.solver.constraints))
+        self.assertEqual(self.n_variables, len(TESTMODEL.solver.variables))
+
+    def test_with(self):
+        with ProblemCache(TESTMODEL) as cache:
+            lmoma(model=TESTMODEL, reference=self.reference, cache=cache)
+            # After running lmoma with cache, the number of variables and constraints are 2 times bigger
+            self.assertEqual(2*self.n_variables, len(TESTMODEL.solver.variables))
```
KristianJensen: Same as above

tests/test_util.py
```diff
@@ -54,6 +61,43 @@ def test_with_statement(self):
...
+            # And has 2 more constraints per reactions
+            self.assertEqual(self.n_constraints + 2 * len(TESTMODEL.reactions), len(TESTMODEL.solver.constraints))
+        # After reset cache, the problem should return to its original size
+        self.assertEqual(self.n_constraints, len(TESTMODEL.solver.constraints))
+        self.assertEqual(self.n_variables, len(TESTMODEL.solver.variables))
+
```
KristianJensen: It would be nice with some unittests that test the API of ProblemCache, i.e. a test for add_constraint, add_variable, etc.

tests/test_util.py
joaocardoso: I agree with you! However, for practical reasons if lmoma fails it will fail on the unit tests for simulation methods. Is a necessary thing or more like an issue that can be fixed later?

KristianJensen: It's not a big issue if you just call it maybe "test_problem_cache_with_lmoma".
There should still be some more unittests for the ProblemCache itself though.

joaocardoso:
More and more tests

PR #77, opened 2016-10-13
Improve cached lmoma (and moma too btw)

joaocardoso:
%timeit lmoma(model, reference=ref, cache=cache)

before: 1 loop, best of 3: 4.08 s per loop
after: 1 loop, best of 3: 681 ms per loop


cameo/flux_analysis/simulation.py
```diff
@@ -178,8 +178,9 @@ def create_constraint(model, constraint_id, var, reaction, flux_value):
                 return constraint
 
             def update_constraint(model, constraint, var, reaction, flux_value):
-                constraint.lb = flux_value
-                constraint.ub = flux_value
+                if constraint.lb != flux_value:
+                    constraint.lb = flux_value
+                    constraint.ub = flux_value
```
joaocardoso: No. It has to be always both. If one is, the other is too (look at the create function). There is no need to test both.

PR #80, opened 2016-10-17
New version of metabolite knock_out keeping the steady-state constraints.

joaocardoso:
Same assumptions as before but now I am adding a sink to the metabolite being knocked out instead of removing relaxing the mass balance constraint.

@KristianJensen, @phantomas1234 I need you to take a fresh look at this please. Two weird things going on:
1. cplex does not raise SolverError any time when I set the lower bound of the objective to 99999.
2. GLPK is returning UndefinedSolution error.

Any thoughts?

### DIFF
--- a/cameo/core/metabolite.py
+++ b/cameo/core/metabolite.py
@@ -11,11 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import logging
 from functools import partial
 
 import cobra
-import logging
 import six
+
 from cameo.util import inheritdocstring
 
 logger = logging.getLogger(__name__)
@@ -54,17 +55,17 @@ class Metabolite(cobra.core.Metabolite):
         else:
             return None
 
-    def _relax_mass_balance_constrain(self, time_machine, absolute_bound):
+    def _relax_mass_balance_constrain(self, time_machine):
         if time_machine:
-            time_machine(do=partial(setattr, self.constraint, "lb", -absolute_bound),
+            time_machine(do=partial(setattr, self.constraint, "lb", -1000 * len(self.reactions)),
                          undo=partial(setattr, self.constraint, "lb", self.constraint.lb))
-            time_machine(do=partial(setattr, self.constraint, "ub", absolute_bound),
+            time_machine(do=partial(setattr, self.constraint, "ub", 1000 * len(self.reactions)),
                          undo=partial(setattr, self.constraint, "ub", self.constraint.ub))
         else:
-            self.constraint.lb = -absolute_bound
-            self.constraint.ub = absolute_bound
+            self.constraint.lb = None
+            self.constraint.ub = None
 
-    def knock_out(self, time_machine=None, absolute_bound=10000, force_steady_state=False):
+    def knock_out(self, time_machine=None, force_steady_state=False):
         """'Knockout' a metabolite. This can be done in 2 ways:
 
         1. Implementation follows the description in [1]
@@ -83,10 +84,7 @@ class Metabolite(cobra.core.Metabolite):
         Parameters
         ----------
         time_machine : TimeMachine
-            An action stack to reverse actions.
-        absolute_bound: number
-            The metabolites is 'knocked-out' by setting the associated constraints in the S-matrix to -absolute_bound
-            <= Si <= absolute_bound so should be a large number to make it effectively unconstrained.
+            An action stack to reverse actions
         force_steady_state: bool
             If True, uses approach 2.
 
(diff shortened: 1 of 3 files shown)


tests/test_solver_based_model.py
```diff
@@ -958,6 +958,21 @@ def test_essential_reactions(self):
...
                 self.model.essential_reactions()
 
+        def test_essential_metabolites(self):
+            model = self.model.copy()
+            essential_metabolites_unbalanced = [m.id for m in model.essential_metabolites(force_steady_state=False)]
+            essential_metabolites_balanced = [m.id for m in model.essential_metabolites(force_steady_state=True)]
+            self.assertTrue(sorted(essential_metabolites_unbalanced) == sorted(ESSENTIAL_METABOLITES))
+            self.assertTrue(sorted(essential_metabolites_balanced) == sorted(ESSENTIAL_METABOLITES))
+
+            with self.assertRaises(cameo.exceptions.SolveError):
```
KristianJensen: With cplex I don't get an exception. I guess I'm not supposed to either?

tests/test_solver_based_model.py
```diff
@@ -958,6 +958,21 @@ def test_essential_reactions(self):
...
+                self.model.reactions.Biomass_Ecoli_core_N_LPAREN_w_FSLASH_GAM_RPAREN__Nmet2.lower_bound = 999999.
+                self.model.essential_metabolites(force_steady_state=False)
+
+            with self.assertRaises(cameo.exceptions.SolveError):
```
KristianJensen: This one craps out with GLPK. It seems to be a(nother) weird issue where GLPK gets stuck with an invalid basis, which results in an undefined solution. It can't get out of this without some help. I think I've fixed it in optlang.

tests/test_solver_based_model.py
joaocardoso: I think you should. I set the lower bound to 99999, which should be Infeasible right?

cameo/core/metabolite.py
```diff
@@ -64,14 +64,20 @@ def _relax_mass_balance_constrain(self, time_machine, absolute_bound):
             self.constraint.lb = -absolute_bound
             self.constraint.ub = absolute_bound
 
-    def knock_out(self, time_machine=None, absolute_bound=10000):
-        """'Knockout' a metabolite
+    def knock_out(self, time_machine=None, absolute_bound=10000, force_steady_state=False):
```
joaocardoso: I wouldn't go with that. Depending on how many reactions your metabolite is present sum(S.v) might become to be a very large number.

tests/test_solver_based_model.py
KristianJensen: Ok, with your latest commit the infeasible is now raised, like essential_reactions would do.

tests/test_solver_based_model.py
phantomas1234: @KristianJensen I'll make a new optlang release that includes your fix. Shouldn't the large bound be determined dynamically from the existing bounds in the model?

joaocardoso: Is there a way to set a bound to None? It should be unbounded

KristianJensen: @joaocardoso Setting lb/ub on optlang variables/constraints to None will make it unbounded in that direction.

joaocardoso: @KristianJensen 
The doesn't hold for CPLEX interface.

1) ERROR: test_knock_out (test_solver_based_model.TestMetaboliteCPLEX)

   Traceback (most recent call last):
    tests/test_solver_based_model.py line 1078 in test_knock_out
      metabolite_a.knock_out(time_machine=tm)
    cameo/core/metabolite.py line 110 in knock_out
      self._relax_mass_balance_constrain(time_machine)
    cameo/core/metabolite.py line 60 in _relax_mass_balance_constrain
      undo=partial(setattr, self.constraint, "lb", self.constraint.lb))
    cameo/util.py line 309 in __call__
      output = do()
    /Users/joao/Documents/repos/optlang/optlang/cplex_interface.py line 264 in lb
      if self.ub is not None and value > self.ub:
   TypeError: unorderable types: NoneType() > int()

KristianJensen: That looks like a bug in optlang. I can fix it tomorrow.

joaocardoso:
@phantomas1234 @KristianJensen now it is just the GLPK undefined $$%#@!@#$

tests/test_solver_based_model.py
KristianJensen: It should be possible to make cplex constraints unconstrained now. Only in one direction though as cplex doesn't support completely free constraints, but you also only need to relax it in the positive direction.

PR #83, opened 2016-11-07
chore: make travis report to slack less frequently

phantomas1234:


hredestig:
any idea why this is failing all of a sudden?

phantomas1234:
Looks like the server in minho is still crapping out from time to time. Also, there is a bug in a new version of optlang. I just merged Kristian's fix and made a new optlang release. Should rerun the travis builds when it available on pypi.

PR #84, opened 2016-11-11
Make use of batch retrieval inside Solution

joaocardoso:


cameo/core/solution.py
```diff
@@ -160,11 +158,17 @@ def __init__(self, model, *args, **kwargs):
         self.fluxes = OrderedDict()
         self.shadow_prices = OrderedDict()
         self.reduced_costs = OrderedDict()
+        self._primal_values = model.solver.primal_values
+        self._dual_values = model.solver.dual_values
```
KristianJensen: This should be `model.solver.shadow_prices`. The function `dual_values` gives the constraint primal values (which doesn't really make sense and will be changed soon)

cameo/core/solution.py
```diff
@@ -250,26 +257,32 @@ def f(self, value):
...
     @property
     def shadow_prices(self):
         self._check_freshness()
-        duals = OrderedDict()
+        dual_values = self.model.solver.dual_values
+        shadow_prices = {}
         for metabolite in self.model.metabolites:
-            duals[metabolite.id] = self.model.solver.constraints[metabolite.id].dual
-        return duals
+            shadow_prices[metabolite.id] = dual_values[metabolite.id]
```
KristianJensen: The loop is unnecessary, it could just directly return the shadow prices from the solver.

cameo/core/solution.py
```diff
@@ -250,26 +257,32 @@ def f(self, value):
...
-            duals[reaction.id] = reaction.reduced_cost
-        return duals
+            reduced_costs[reaction.id] = reduced_values[reaction._get_forward_id()] - reduced_values[reaction._get_reverse_id()]
+        return reduced_costs
 
     @property
     def shadow_prices(self):
         self._check_freshness()
-        duals = OrderedDict()
+        dual_values = self.model.solver.dual_values
```
KristianJensen: This should also be `self.model.solver.shadow_prices`

hredestig:
@KristianJensen @joaocardoso 
I spotted this after merge, can you reproduce and if so fix?
```
>>> from cameo import load_model
>>> e = load_model('tests/data/EcoliCore.xml')
>>> s = e.solve()
>>> s.reduced_costs
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/henred/proj/cameo/cameo/core/solution.py", line 271, in reduced_costs
    reduced_costs[reaction.id] = reduced_values[reaction._get_forward_id()] - reduced_values[reaction._get_reverse_id()]
KeyError
```

joaocardoso:
@hredestig can you try to update optlang? I think @KristianJensen changed some stuff in there for this to work.

hredestig:
alright, indeed.. seems we are missing a test here, should not have been caught by CI which uses released optlang

PR #85, opened 2016-11-11
High level api

joaocardoso:
Hi @biosustain/team-modeling 
The final effort to get the high-level API running smoothly is here.

#### Please read everything before asking questions that are might be answered bellow! Thanks

Before any changes, things work like this: methods on strain_design are subclass of StrainDesignMethod. A StrainDesignMethod.run returns a StrainDesignMethodResult, which contains a list of StrainDesign. The StrainDesign contains a DictList of Target (Target is explained bellow).

Added stuff:
* Target (anything from GeneManipulationTarget, GeneKnockoutTarget, ReactionCofactorSwapTarget, ReactionKnockinTarget, etc.) basically everything that changes the model and is predicted by a strain design method.
* swap_cofators method to the Reaction (@hredestig I told you so before).
* DifferentialFVA to the high-level API (because the results are now comparable and 'aggregatable')
* OptSwap as a StrainDesignMethod (like OptGene). Let's make it clear that stuff under cameo.strain_design.heuristic.evolutionary is low-level API, not to be exposed as StrainDesignMethod.
* Target output gnomic if python.version >= 3.

Refactored stuff:
* Evaluators moved to cameo.strain_design.heuristic.evolutionary.evaluators because cameo.strain_design.heuristic.optimization became a mess.
* cameo.strain_design.heuristic.decoders got more generic names for what they do.
* since Gene and Reaction support knock_out() there is no need for weird stuff coming out of the decoders any more.
* FSEOF, DifferentialFVA, OptKnock, OptGene, PathwayPredictor and OptSwap are all subclass of StrainDesignMethod and all outputs are converted into target.
* KnockoutOptimziation was renamed to TargetOptimization and only specific stuff is implemented.
* KnockoutOptimizationResult was renamed to TargetOptimizationResult and has a metadata field to include specific stuff from implementations of TargetOptimization (such as cofactors to swap).


cameo/core/reaction.py
```diff
@@ -448,6 +448,26 @@ def _(reaction, lb, ub):
         else:
             super(Reaction, self).knock_out()
 
+    def swap_cofactors(self, swap_pairs, time_machine=None):
```
KristianJensen: This needs a docstring

cameo/core/reaction.py
```diff
@@ -448,6 +448,26 @@ def _(reaction, lb, ub):
...
+        if all(self.metabolites.get(met, False) for met in swap_pairs[0]):
```
KristianJensen: Maybe use 'all(met in self.metabolites)'? It's a bit more explicit

cameo/strain_design/core.py
```diff
@@ -0,0 +1,319 @@
...
+from pandas import DataFrame
+
+from cameo import ui
+from cameo.core.result import Result
+
+try:
+    from gnomic import Genotype
+    from gnomic.models import Feature, Accession, Del, Ins, Sub, Mutation
+    _gnomic_available = True
+except Exception:
```
KristianJensen: Any specific reason this is not except ImportError?

cameo/strain_design/core.py
```diff
@@ -0,0 +1,319 @@
...
+    _gnomic_available = False
+
+
+class Target(object):
+    def __init__(self, id):
+        self.id = id
+
+    def apply(self, model, time_machine=None):
+        raise NotImplemented
```
KristianJensen: This should be raise NotImplementedError

cameo/strain_design/core.py
```diff
@@ -0,0 +1,319 @@
...
+class FluxModulationTarget(Target):
+    __gnomic_feature_type__ = 'flux'
+
+    def __init__(self, id, value, reference_value):
+        super(FluxModulationTarget, self).__init__(id)
+        self._reference_value = reference_value
+        self._value = value
+
+    def get_target(self, model):
+        raise NotImplemented
```
KristianJensen: Should be raise NotImplementedError

cameo/strain_design/core.py
```diff
@@ -0,0 +1,319 @@
...
+            over_expression = Feature(accession=accession, type=self.__gnomic_feature_type__,
+                                      variant="over-expression(%f)" % self.fold_change)
+            return Mutation(feature, over_expression)
+        elif self._value < self._reference_value:
+            under_expression = Feature(accession=accession, type=self.__gnomic_feature_type__,
+                                       variant="down-regulation(%f)" % self.fold_change)
+            return Mutation(feature, under_expression)
+
+
+class SwapTarget(Target):
```
KristianJensen: Is this unfinished? It looks like it's not used

cameo/strain_design/core.py
```diff
@@ -0,0 +1,319 @@
+# Copyright 2015 Novo Nordisk Foundation Center for Biosustainability, DTU.
```
KristianJensen: There are no docstrings in this module, for classes or methods. Adding some would make it more clear what the different classes are meant to be used for.

cameo/strain_design/core.py
```diff
@@ -0,0 +1,319 @@
...
+            else:
+                return all(self.targets[i] == other.targets[i] for i in range(self))
+        else:
+            return False
+
+    def apply(self, model, time_machine=None):
+        for target in self.targets:
+            target.apply(model, time_machine)
+
+    def __add__(self, other):
```
KristianJensen: I think this will raise an error if the two strain designs have one or more targets in common. Maybe use the DictList.union instead?

cameo/strain_design/core.py
```diff
@@ -0,0 +1,319 @@
...
+        """
+        Return the number of designs found.
+        """
+        return len(self._designs)
+
+    def __iter__(self):
+        """
+        Returns an iterator that yields StrainDesign objects.
+        """
+        for design in self._designs:
```
KristianJensen: Could also just be `return iter(self._designs)`

cameo/strain_design/heuristic/evolutionary/evaluators.py
```diff
@@ -0,0 +1,187 @@
...
+from cameo.util import ProblemCache, memoize, TimeMachine
+
+logger = logging.getLogger(__name__)
+
+__all__ = ['KnockoutEvaluator', 'SwapEvaluator']
+
+
+class Evaluator(object):
+    def __call__(self, population):
+        raise NotImplemented
```
KristianJensen: NotImplementedError

cameo/core/reaction.py
joaocardoso: True, that makes more sense actually.

cameo/strain_design/core.py
joaocardoso: This is a tricky one! Some targets will be complementary and others will not be compatible:
Compatible: add reaction X and swap cofactors
Not compatible: over-express reaction Y and knockout reaction Y.

Any suggestions?

cameo/strain_design/core.py
joaocardoso: Because in PY2 it raises a SyntaxError

joaocardoso:
@KristianJensen: I made it possible to merge Target. If targets are incompatible it will raise an exception.

phantomas1234:
image

The progress bar for the optimizations ends up at the top. Can it be moved to the bottom of the running computation?

Midnighter:
Not sure what you're currently using for the progressbar. I've been quite impressed with [tqdm](https://github.com/tqdm/tqdm) so far.

cameo/strain_design/core.py
joaocardoso: I mean, it doesn't raise an ImportError, it raises a SyntaxError.
Here:
https://travis-ci.org/biosustain/cameo/jobs/174900373#L1072

cameo/strain_design/core.py
```diff
@@ -0,0 +1,514 @@
...
+            over_expression = Feature(accession=accession, type=self.__gnomic_feature_type__,
+                                      variant="over-expression(%f)" % self.fold_change)
+            return Mutation(feature, over_expression)
+        elif self._value < self._reference_value:
+            under_expression = Feature(accession=accession, type=self.__gnomic_feature_type__,
+                                       variant="down-regulation(%f)" % self.fold_change)
+            return Mutation(feature, under_expression)
+
+
+class ReactionCofactorSwapTarget(Target):
```
joaocardoso: My opinion is that adding and removing variables is slower than changing coefficients. If I want to see what happens if the same reaction had a different cofactor specificity I would write:
reaction.add_metabolites({nad:-1, nadp:1, nadh:-1, napdh1:}) instead of adding and removing reactions. In the other hand I there are situations where we might want to test both reactions with different cofactors with different ratios, etc...

I'll cook something to solve this problem.

cameo/strain_design/core.py
phantomas1234: Also move this into `cameo.core.target` and `cameo.core.strain_design`

phantomas1234:
I am going to merge this now, come hell and high water ...

cameo/strain_design/heuristic/evolutionary/evaluators.py
```diff
@@ -0,0 +1,191 @@
...
+        self.cache = ProblemCache(model)
+
+    def __call__(self, population):
+        return [self._evaluate_individual(tuple(i)) for i in population]
+
+    def reset(self):
+        self.cache.reset()
+
+    @property
+    def is_mo(self):
```
hredestig: has_multiobjective?

cameo/strain_design/heuristic/evolutionary/evaluators.py
```diff
@@ -0,0 +1,191 @@
...
+
+class KnockoutEvaluator(TargetEvaluator):
+    """
+    Knockout evaluator for genes or reactions.
+    """
+
+    @memoize
+    def _evaluate_individual(self, individual):
+        """
+        Evaluates a single individual.
```
hredestig: rephrase, elaborate

cameo/strain_design/heuristic/evolutionary/processing.py
```diff
@@ -13,12 +13,113 @@
...
 from cameo import flux_variability_analysis
 from cameo.util import TimeMachine
 
 
-def process_knockout_solution(model, solution, simulation_method, simulation_kwargs,
-                              biomass, target, substrate, objective_functions):
+def process_reaction_knockout_solution(model, solution, simulation_method, simulation_kwargs,
+                                       biomass, target, substrate, objective_functions):
+    """
+
```
hredestig: doc

cameo/api/designer.py
```diff
@@ -20,6 +20,7 @@
 import re
 
 import numpy as np
+from six.moves import reduce
```
hredestig: functools?

PR #89, opened 2016-12-01
fix: drop (most) cobrapy tests

hredestig:
cobrapy now uses pytest so the direct import to cameo is not
straight-forward anymore. since this kind of compatiblity won't be
needed in the future I suggest not spending time solving this properly.


phantomas1234:
Makes sense

PR #91, opened 2016-12-16
fix: remove non-existent libglp dependency

phantomas1234:
travis-ci used to ignore it in earlier builds


phantomas1234:
@joaocardoso actually it was me who added this typo long time ago but travis-ci always ignored it.

PR #93, opened 2016-12-21
Feature: Implement multiobjective function

joaocardoso:
Highlights:
* Retrieve worst_fitness from objective functions
* Remove is_mo method
* No lists as objective functions
* Fix and improve unit tests

More info:
This solves the problem of is_mo property or checking if there is a list of objectives.


joaocardoso:
I found a little bit more stuff to fix meanwhile.

PR #95, opened 2017-01-05
Fix plotting stuff

joaocardoso:
I fixed the labels on the phenotypic phase plane. 

* If it the reaction is biomass, then the unit becomes h-1.
* In 3d, the objective unit is the estimate, but the variables are still fluxes.


cameo/flux_analysis/analysis.py
```diff
@@ -676,6 +676,13 @@ def plot(self, grid=None, width=None, height=None, title=None, axis_font_size=No
         if grid is None:
             plotter.display(plot)
 
+    @staticmethod
+    def _axis_label(variable_id, nice_variable_id, unit):
+        if "BIOMASS" in variable_id:
```
hredestig: obviously won't always work but it's an improvement - make case insensitive to cover a couple of more situations?

PR #106, opened 2017-01-30
Flux inversion targets

joaocardoso:
A target to repren

### DIFF
--- a/cameo/core/manipulation.py
+++ b/cameo/core/manipulation.py
@@ -84,6 +84,36 @@ def decrease_flux(reaction, ref_value, value, time_machine=None):
         reaction.knock_out(time_machine=time_machine)
 
 
+def reverse_flux(reaction, ref_value, value, time_machine=None):
+    """
+    lb                           0                           ub
+    |--------------------------- ' ---------------------------|
+                      <----------'
+                                 '---------->
+
+    Parameters
+    ----------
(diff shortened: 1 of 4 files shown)


cameo/core/manipulation.py
```diff
@@ -84,6 +84,36 @@ def decrease_flux(reaction, ref_value, value, time_machine=None):
         reaction.knock_out(time_machine=time_machine)
 
 
+def reverse_flux(reaction, ref_value, value, time_machine=None):
+    """
+    lb                           0                           ub
```
hredestig: wouldn't hurt with a couple of words like 'change the direction of the bounds on flux with respect to a reference value' or something like that. the ascii diagram isn't super intuitive...

cameo/core/manipulation.py
```diff
@@ -84,6 +84,36 @@ def decrease_flux(reaction, ref_value, value, time_machine=None):
...
+    |--------------------------- ' ---------------------------|
+                      <----------'
+                                 '---------->
+
+    Parameters
+    ----------
+    reaction: cameo.Reaction
+        The reaction to down_regulate.
```
hredestig: not necessarily a 'down regulation'?

joaocardoso: It's actually not a down regulation at all 👍 
I'll fix that

PR #108, opened 2017-02-01
Feature: support multiple substrates on yield functions

joaocardoso:


cameo/strain_design/heuristic/evolutionary/objective_functions.py
```diff
@@ -116,18 +116,25 @@ def _repr_latex_(self):
 
 
 class YieldFunction(ObjectiveFunction):
```
KristianJensen: This could do with a docstring, e.g. specifying the allowed types for the arguments

cameo/strain_design/heuristic/evolutionary/objective_functions.py
```diff
@@ -116,18 +116,25 @@ def _repr_latex_(self):
...
-    def __init__(self, product, substrate, *args, **kwargs):
+    def __init__(self, product, substrates, *args, **kwargs):
         super(YieldFunction, self).__init__(*args, **kwargs)
-        self.n_c_substrate = 1
+
+        self.n_c_substrates = {}
         self.n_c_product = 1
         if isinstance(product, Reaction):
```
KristianJensen: Maybe a check that the reaction is an exchange reaction before just taking the 0th reactant

KristianJensen: What other types is `product` allowed to be? Metabolite?

cameo/strain_design/heuristic/evolutionary/objective_functions.py
```diff
@@ -116,18 +116,25 @@ def _repr_latex_(self):
...
             self.n_c_product = product.reactants[0].elements.get('C', 1)
             product = product.id
         self.product = product
-        if isinstance(substrate, Reaction):
-            self.n_c_substrate = substrate.reactants[0].elements.get('C', 1)
-            substrate = substrate.id
-        self.substrate = substrate
+
+        if not isinstance(substrates, list):
```
KristianJensen: This is a bit dangerous. The variable could be a tuple, series or similar. It would be better to enumerate the types that are allowed as a "single substrate" argument, e.g. `if isinstance(substrates, (six.string_types, Reaction))`. Another possibility would be to force the user to actually give a sequence type object (the argument is called substrateS after all)

cameo/strain_design/heuristic/evolutionary/objective_functions.py
```diff
@@ -116,18 +116,25 @@ def _repr_latex_(self):
...
+            substrates = [substrates]
+
+        for i, substrate in enumerate(substrates):
+            if isinstance(substrate, Reaction):
```
KristianJensen: What else can it be?

cameo/strain_design/heuristic/evolutionary/objective_functions.py
```diff
@@ -165,36 +172,48 @@ class biomass_product_coupled_yield(YieldFunction):
...
         self.biomass = biomass
+        self.carbon_yield = carbon_yield
 
     def __call__(self, model, solution, targets):
-        try:
-            biomass_flux = round(solution.fluxes[self.biomass], config.ndecimals)
+        biomass_flux = round(solution.fluxes[self.biomass], config.ndecimals)
+        if self.carbon_yield:
+            product_flux = round(solution.fluxes[self.product], config.ndecimals) * self.n_c_product
+            substrate_flux = sum(round(abs(solution.fluxes[s]), config.ndecimals) * self.n_c_substrates.get(s, 1)
```
KristianJensen: You shouldn't use get here I think. If the substrate is not in the dictionary there should be an error.

KristianJensen: Taking the absolute value can in theory give some weird results. If one substrate is suddenly produced for some reason it will still count toward the uptake flux.

cameo/strain_design/heuristic/evolutionary/objective_functions.py
```diff
@@ -165,36 +172,48 @@ class biomass_product_coupled_yield(YieldFunction):
...
+                                 for s in self.substrates)
+
+        else:
             product_flux = round(solution.fluxes[self.product], config.ndecimals)
-            substrate_flux = round(abs(solution.fluxes[self.substrate]), config.ndecimals)
+            substrate_flux = sum(round(abs(solution.fluxes[s]), config.ndecimals) for s in self.substrates)
+        try:
             return round((biomass_flux * product_flux) / substrate_flux, config.ndecimals)
```
KristianJensen: Rounding is not necessary since all the fluxes are already rounded.

cameo/strain_design/heuristic/evolutionary/objective_functions.py
```diff
@@ -214,35 +233,35 @@ class biomass_product_coupled_min_yield(YieldFunction):
...
     def __call__(self, model, solution, targets):
-        try:
-            biomass_flux = round(solution.fluxes[self.biomass], config.ndecimals)
-            fva_res = flux_variability_analysis(model, reactions=[self.product], fraction_of_optimum=1)
-            min_product_flux = fva_res["lower_bound"][self.product]
-
-            substrate_flux = round(abs(solution.fluxes[self.substrate]), config.ndecimals)
-            return round((biomass_flux * min_product_flux) / substrate_flux, config.ndecimals)
+        biomass_flux = round(solution.fluxes[self.biomass], config.ndecimals)
+        fva_res = flux_variability_analysis(model, reactions=[self.product], fraction_of_optimum=1)
```
KristianJensen: This will solve the model twice (min and max). It could be made faster by manually setting the objective and only solving the minimization

cameo/strain_design/heuristic/evolutionary/objective_functions.py
```diff
@@ -214,35 +233,35 @@ class biomass_product_coupled_min_yield(YieldFunction):
...
+        min_product_flux = round(fva_res["lower_bound"][self.product], config.ndecimals)
+        if self.carbon_yield:
+            product_flux = min_product_flux * self.n_c_product
+            substrate_flux = sum(round(abs(solution.fluxes[s]), config.ndecimals) * self.n_c_substrates.get(s, 1)
```
KristianJensen: Same as above, get should not be necessary

cameo/strain_design/heuristic/evolutionary/objective_functions.py
joaocardoso: That can be implemented later if need be

cameo/core/reaction.py
```diff
@@ -379,6 +379,10 @@ def reduced_cost(self):
         else:
             return None
 
+    @property
+    def is_exchange(self):
+        return len(self.reactants) == 0 or len(self.products) == 0
```
KristianJensen: Check for only one metabolite

cameo/strain_design/heuristic/evolutionary/objective_functions.py
KristianJensen: This should be changed to check for string or Reaction instead. This way other iterables than list will work

PR #119, opened 2017-02-27
Nullspace simplification

joaocardoso:
Me and @KristianJensen use nullspace simplification to make methods such as OptGene and OptKnock faster by reducing the search space.

Credit to @KristianJensen for implementing all the utils, dead end and reaction groups needed to make this easy to implement.


phantomas1234:
@Midnighter and/or @hredestig can you please review this

cameo/flux_analysis/structural.py
```diff
@@ -16,25 +16,118 @@
...
+        nullspace of `A`.  The columns of `ns` are a basis for the
+        nullspace; each element in numpy.dot(A, ns) will be approximately
+        zero.
+    """
+
+    matrix = np.atleast_2d(matrix)
+    u, s, vh = svd(matrix)
+    tol = max(atol, rtol * s[0])
+    nnz = (s >= tol).sum()
+    ns = vh[nnz:].conj().T
```
Midnighter: I find the following more legible but they are equivalent:
```Python
ns = np.compress(s <= tol, vh, axis=0)
```

cameo/flux_analysis/structural.py
```diff
@@ -16,25 +16,118 @@
...
+    return ns
+
+
+def find_blocked_reactions_nullspace(model, ns=None, tol=1e-10):
+    """Identify reactions that can't carry flux based on the nullspace of the stoichiometric matrix.
+    A blocked reaction will correspond to an all-zero row in N(S)."""
+    if ns is None:
+        ns = nullspace(model.S)
+    mask = (np.abs(ns) <= tol).all(1)
+    blocked = [reac for reac, b in zip(model.reactions, mask) if b is True]
```
Midnighter: If you use the `future` package. You can `from builtins import zip` which will give you a version of `zip` in Python2 & 3 that works like `itertools.izip`, i.e., does not create a list first.

cameo/flux_analysis/structural.py
```diff
@@ -16,25 +16,118 @@
...
+    """
+    Find groups of reactions whose fluxes are forced to be multiples of each other.
+    """
+    if ns is None:
+        ns = nullspace(model.S)
+    blocked = set(find_blocked_reactions_nullspace(model, ns, tol))
+    blocked_mask = np.array([True if r in blocked else False for r in model.reactions])
+    non_blocked_ns = ns[~blocked_mask]
+    non_blocked_reactions = np.array(list(model.reactions))[~blocked_mask]
+
```
Midnighter: Seems less convoluted to pull in code from `find_blocked_reactions_nullspace` in here. Compute the mask directly on the nullspace as above `mask = (np.abs(ns) <= tol).all(axis=1)`, find blocked reactions, and non-blocked are `set(reactions) - set(blocked)`.

cameo/flux_analysis/structural.py
```diff
@@ -16,25 +16,118 @@
...
+    # Calculate correlation coefficients (perfect correlation means that one is a scalar multiple of the other)
+    corr_mat = np.corrcoef(non_blocked_ns)
+    dist_mat = 1 - np.abs(corr_mat)
+
+    # Perform a linkage clustering
+    link = linkage(dist_mat, method="complete")
+
+    # Filter the linkage results (only distance = 0)
+    good_link = [a for a in link if a[2] < tol]
```
Midnighter: `good_link = link[link[:, 2] <= tol]`?

cameo/flux_analysis/structural.py
```diff
@@ -16,25 +16,118 @@
...
+
+    # Aggregate groups
+    group_dict = {i: {i} for i in range(len(non_blocked_reactions))}
+    for i, n in enumerate(range(len(non_blocked_reactions), len(non_blocked_reactions) + len(good_link))):
+        a = good_link[i]
+        group_dict[n] = group_dict[a[0]] | group_dict[a[1]]
+        del group_dict[a[0]]
+        del group_dict[a[1]]
+
+    groups = [set(non_blocked_reactions[np.array(list(v))]) for v in group_dict.values()]
```
Midnighter: Maybe use `six` or `future` here to iterate over values in Python 2 and not create a list.

Midnighter: You can also index with a list directly into an array rather than creating an array.

Midnighter: And combine the below line so you only iterate once, i.e., `if len(v) > 1`.

cameo/strain_design/heuristic/evolutionary/optimization.py
```diff
@@ -573,24 +577,37 @@ class ReactionKnockoutOptimization(KnockoutOptimization):
...
         else:
             self.essential_reactions = set([r.id for r in self.model.essential_reactions()] + essential_reactions)
 
-        exchange_reactions = set([r.id for r in self.model.exchanges])
-        self.representation = list(self.reactions.difference(self.essential_reactions).difference(exchange_reactions))
+        if use_nullspace_simplification:
+            ns = nullspace(self.model.S)
+            dead_ends = find_blocked_reactions_nullspace(self.model, ns=ns)
+            exchanges = self.model.exchanges
+            reactions = [r for r in self.model.reactions if r not in exchanges and r not in dead_ends and
```
Midnighter: Would be better to ensure that `exchanges`, `dead_ends`, and `self.essential_reactions` are all sets. I don't think `dead_ends` and `exchanges` are giving this line potentially terrible performance.

cameo/strain_design/heuristic/evolutionary/optimization.py
```diff
@@ -573,24 +577,37 @@ class ReactionKnockoutOptimization(KnockoutOptimization):
...
+                         r.id not in self.essential_reactions]
+
+            groups = find_coupled_reactions_nullspace(self.model, ns=ns)
+            groups = [group for group in groups if any(r.id in reactions for r in group)]
+            reduced_set = reduce_reaction_set(reactions, groups)
+            to_keep = [r.id for r in reduced_set]
```
Midnighter: Can't see from the code shown here if it matters down the line, but maybe make `to_keep` a set, too.

cameo/strain_design/heuristic/evolutionary/optimization.py
```diff
@@ -663,7 +682,17 @@ def __init__(self, genes=None, essential_genes=None, *args, **kwargs):
...
             self.essential_genes = set([g.id for g in self.model.essential_genes()] + essential_genes)
 
-        self.representation = list(self.genes.difference(self.essential_genes))
+        # TODO: use genes from groups
+        if use_nullspace_simplification:
+            ns = nullspace(self.model.S)
+            dead_end_reactions = find_blocked_reactions_nullspace(self.model, ns=ns)
+            dead_end_genes = [g for g in self.model.genes if all(r in dead_end_reactions for r in g.reaction)]
+            genes = [g for g in self.model.genes if g not in self.essential_genes and g.id not in dead_end_genes]
+            self.representation = [g.id for g in genes]
```
Midnighter: Same, as comment above, use sets for much better `__contains__` performance.

cameo/strain_design/heuristic/evolutionary/optimization.py
joaocardoso: Hi @Midnighter! `to_keep` is going to be assigned to `self.representation` must be a list because it needs to be indexed.

cameo/flux_analysis/structural.py
KristianJensen: There should not be any complex values for this use case. The function finds the nullspace in the general case as it comes directly from a scipy example.

cameo/flux_analysis/structural.py
KristianJensen: I personally find `if b is True` to be more readable. The performance difference is about as negligible as it gets.

cameo/flux_analysis/structural.py
KristianJensen: I get your point. I wanted to avoid code duplication but in this case you're right, it does seem to get overly complicated.

KristianJensen:
@Midnighter You're right it does need a couple of unittests. I've pushed some changes wrt the comments you had.

Midnighter:
Looks good so far. I can approve now or wait and take a look at the tests. Whichever you prefer.

joaocardoso:
I will add some tests now.

joaocardoso:
@Midnighter we have tests now. I think we are good to go.

cameo/core/solver_based_model.py
```diff
@@ -634,6 +635,19 @@ def essential_genes(self, threshold=1e-6):
         return essential
 
     @property
+    def S(self):
```
hredestig: suggesting to drop this property in favor of a function `create_stoichiometric_array` as that will make cobrapy merge easier

cameo/util.py
```diff
@@ -608,3 +565,38 @@ def zip_repeat(long_iter, short_iter):
     """
     for i, j in zip(long_iter, itertools.cycle(short_iter)):
         yield i, j
+
+
+def pick_one(iterable):
```
hredestig: I find `next(iter(x))` more readable than defining new function. `pick_one`... which one? any? :)

PR #123, opened 2017-03-06
fix: fva flags

joaocardoso:
Now that it runs by getting all the `min` and then all the `max`
the lower_bound flags need to be stored and evaluated in the second iteration.
I added a dictionary to store the lb_flag.


cameo/flux_analysis/analysis.py
```diff
@@ -226,24 +226,31 @@ def _flux_variability_analysis(model, reactions=None):
...
+            model.solver.objective.set_linear_coefficients({reaction.forward_variable: 0.,
+                                                            reaction.reverse_variable: 0.})
+
+            assert model.objective.expression == 0, model.objective.expression
 
         for reaction in reactions:
-            model.solver.objective.set_linear_coefficients({reaction.forward_variable: 1., reaction.reverse_variable: -1.})
+            ub_flag = False
+            model.solver.objective.set_linear_coefficients({reaction.forward_variable: 1.,
+                                                            reaction.reverse_variable: -1.})
```
KristianJensen: `model.objective.direction = 'max'` could be moved outside the loop for a small performance boost

KristianJensen:
@phantomas1234 Do you remember the reason for adding the infeasibility flags in FVA in the first place? When just changing the objective as we do in FVA the feasibility of the problem should never change, so I don't really understand the need to catch infeasible exceptions. Is it some numerical issues?

PR #125, opened 2017-03-06
Fix products

### LINKED ISSUE #122: MetaNetX database
The search in the database is not very reliable. I already changed it to be lower-case only, which helped. Also, MetaNetX has mapping for things that are redundant, however I can only find one source and one name, where there should be sources and names or synonyms. 

I cannot find L-Serine, L-Lactate, etc.
There are several algorithms to do this:

* Hamming distance
* Levenshtein distance
* Damerau–Levenshtein distance
* Jaro–Winkler distance

joaocardoso:
Make search case insensitive.

Fixes #122

### DIFF
--- a/cameo/api/products.py
+++ b/cameo/api/products.py
@@ -73,8 +73,11 @@ class Products(object):
             raise Exception("No compound matches found for query %s" % query)
 
     def _search_by_name_fuzzy(self, name):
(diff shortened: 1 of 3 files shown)


joaocardoso:
I have to make this pull request from `deve`l.

PR #139, opened 2017-03-15
Update designer workflow

joaocardoso:
Depends on #138, #136 and #137


tests/test_api.py
```diff
@@ -18,9 +18,29 @@
...
 
-from cameo import api
+from cameo import api, load_model
+from cameo import models, config
+from cameo.api.hosts import Host
 from cameo.api.products import Compound
 
+MODELS = os.path.dirname(models.__file__)
+
+UNIVERSALMODEL = load_model(os.path.join(MODELS, 'json/iJO1366.json'))
```
hredestig: can use the `universal_model` fixture

joaocardoso: good catch! Fixing it already.

PR #141, opened 2017-03-27
fix: designer workflow

joaocardoso:
Less iterations and less pathways.
Fail gracefully if the compound is in products but not in the database

### DIFF
--- a/cameo/api/designer.py
+++ b/cameo/api/designer.py
@@ -359,7 +359,7 @@ class Designer(object):
             search_result = products.search(product)
             search_result = search_result.loc[[i for i in search_result.index if i in database.metabolites]]
             if len(search_result) == 0:
-                raise ("No compound matches found for query %s" % product)
+                raise KeyError("No compound matches found for query %s" % product)
             notice("Found %d compounds that match query '%s'" % (len(search_result), product))
             self.__display_product_search_result(search_result)
             notice("Choosing best match (%s) ... please interrupt if this is not the desired compound."
--- a/cameo/api/designer.py
+++ b/cameo/api/designer.py
@@ -73,7 +73,7 @@ class _OptimizationRunner(object):
 
 class _OptGeneRunner(_OptimizationRunner):
     def __call__(self, strategy):
-        max_evaluations = 20000
+        max_evaluations = 15000
 
         if self.debug:
             max_evaluations = 1000
@@ -98,7 +98,7 @@ class _OptGeneRunner(_OptimizationRunner):
 
 class _DifferentialFVARunner(_OptimizationRunner):
     def __call__(self, strategy):
-        points = 50
+        points = 30
         surface_only = False
         if self.debug:
             points = 5
@@ -170,7 +170,7 @@ class Designer(object):
         notice("Starting searching for compound %s" % product)
         try:
             product = self.__translate_product_to_universal_reactions_model_metabolite(product, database)
-        except KeyError:
+        except Exception:
             raise KeyError("Product %s is not in the %s database" % (product, database.id))
         pathways = self.predict_pathways(product, hosts=hosts, database=database, aerobic=aerobic)
         optimization_reports = self.optimize_strains(pathways, view, aerobic=aerobic)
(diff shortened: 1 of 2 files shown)


cameo/api/designer.py
```diff
@@ -98,7 +98,7 @@ def __call__(self, strategy):
 
 class _DifferentialFVARunner(_OptimizationRunner):
     def __call__(self, strategy):
-        points = 50
+        points = 30
```
hredestig: same as above

cameo/api/designer.py
```diff
@@ -73,7 +73,7 @@ def __call__(self, strategy):
 
 class _OptGeneRunner(_OptimizationRunner):
     def __call__(self, strategy):
-        max_evaluations = 20000
+        max_evaluations = 15000
```
hredestig: you don't want to make this an argument instead? Changing like this seems cumbersome?

cameo/api/designer.py
```diff
@@ -170,7 +170,7 @@ def __call__(self, product='L-glutamate', hosts=HOSTS, database=None, aerobic=Tr
         notice("Starting searching for compound %s" % product)
         try:
             product = self.__translate_product_to_universal_reactions_model_metabolite(product, database)
-        except KeyError:
+        except Exception:
```
hredestig: maybe first catch the KeyError then the catch-all and re-raise it with with some added info rather than converting everything to KeyError?

cameo/api/designer.py
```diff
@@ -317,7 +317,7 @@ def predict_pathways(self, product, hosts=None, database=None, aerobic=True):
         dict
             ([Host, Model] -> PredictedPathways)
         """
-        max_predictions = 8
+        max_predictions = 4
```
hredestig: argument

cameo/api/designer.py
```diff
@@ -357,11 +357,14 @@ def __translate_product_to_universal_reactions_model_metabolite(self, product, d
             return product
         elif isinstance(product, str):
             search_result = products.search(product)
+            search_result = search_result.loc[[i for i in search_result.index if i in database.metabolites]]
+            if len(search_result) == 0:
+                raise ("No compound matches found for query %s" % product)
```
hredestig: KeyError?

cameo/cli/controllers.py
```diff
@@ -108,14 +108,17 @@ def default(self):
...
-                         view=view, aerobic=not self.app.pargs.anaerobic)
+            results['heterologous_pathway'] = results.heterologous_pathway.apply(str)
+            results['manipulations'] = results.manipulations.apply(str)
 
-        results['heterologous_pathway'] = results.heterologous_pathway.apply(str)
-        results['manipulations'] = results.manipulations.apply(str)
-
-        OUTPUT_WRITER[output_format](results, output)
+            OUTPUT_WRITER[output_format](results, output)
+        except KeyError as e:
```
hredestig: I thought print stack-trace and exit(1) was default for uncaught exceptions anyway..?

cameo/api/designer.py
joaocardoso: Yes, good caught. Working on the fix.

cameo/cli/controllers.py
joaocardoso: Not here, at this point we know what raised KeyError there, because we raised it. The CLI should fail gracefully without stack trace for exceptions, just a message explaining the user  why it didn't work.

cameo/api/designer.py
joaocardoso: I agree with you. The next feature is going to be able to give a configuration file with this things. However, for the sake of publication, this is has much as I decided to change.

cameo/api/designer.py
joaocardoso: Yes, I should be able to catch KeyError there. Unfortunately the underlaying implementation needs to be corrected (e.g. `cameo.api.products.products.search`) but there are many other exceptions that need to be refactored. In this case, any exception that comes from `__translate_product_to_universal_reactions_model_metabolite` result from not being able to find any metabolite (could also be index errors on pandas data frames).

joaocardoso: We might even need to create our own exceptions for some of these things.

PR #151, opened 2017-04-03
feat: add a time limit to heuristic optimization

joaocardoso:
Also added fix documentation and unit tests.

### DIFF
--- a/cameo/api/designer.py
+++ b/cameo/api/designer.py
@@ -74,9 +74,11 @@ class _OptimizationRunner(object):
 class _OptGeneRunner(_OptimizationRunner):
     def __call__(self, strategy):
         max_evaluations = 15000
+        max_time = (45, 0)
 
         if self.debug:
             max_evaluations = 1000
+            max_time = (5, 0)
 
         (model, pathway, aerobic) = (strategy[1], strategy[2], strategy[3])
         model = model.copy()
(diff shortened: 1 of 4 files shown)


cameo/strain_design/heuristic/evolutionary/optimization.py
```diff
@@ -205,7 +229,21 @@ def run(self, evaluator=None, generator=None, view=config.default_view, maximize
 
         for observer in self.observers:
             observer.reset()
+
         t = time.time()
+
+        if max_time is not None:
+            terminator = self.heuristic_method.terminator
```
phantomas1234: What if `self.heuristic_method.terminator` is already `inspyred.ec.terminators.time_termination`?

cameo/api/designer.py
```diff
@@ -74,9 +74,11 @@ def __call__(self, strategy):
 class _OptGeneRunner(_OptimizationRunner):
     def __call__(self, strategy):
         max_evaluations = 15000
```
phantomas1234: All of this should not be hard-coded, but let's refactor this later.

cameo/strain_design/heuristic/evolutionary/optimization.py
joaocardoso: then they will both trigger termination with the same max_time so no problem.

cameo/api/designer.py
joaocardoso: I agree, but it needs a general configuration strategy with default values that can be easily defined without using 50 keyword arguments.

PR #152, opened 2017-04-04
[WIP] Make PathwayPredictor use user-provided mapping

phantomas1234:


phantomas1234:
I think I will make those changes against the new branch.

PR #154, opened 2017-04-05
refactor: replace cameo.SolveError with cobra.OptimizationError

phantomas1234:
If we have to refactor all old cameo code to use `optimize` instead of `solve` anyways then there is not point to keep the `SolveError` exception around.


phantomas1234:
@hredestig I rebased this branch onto devel06, so I think this one should be good to go if the tests pass.

PR #156, opened 2017-04-06
Fix/metanetx parsing

phantomas1234:
The different universal models were previously based on the source
column of the reac_prop table from metanetx. Unfortunately, this
did not reflect the actual reaction content of the different databases
(bigg, rhea, ...). Now the universal models are based on the reac_xref
mapping.


scripts/parse_metanetx.py
```diff
@@ -64,13 +64,14 @@ def parse_coeff_and_metabolites(term):
     return stoichiometry
 
 
-def construct_universal_model(list_of_db_prefixes):
+def construct_universal_model(list_of_db_prefixes, reac_xref, reac_prop, chem_prop):
```
hredestig: perhaps it could afford a docstring as not immediately obvious what `reac_xref` should be..

matthiaskoenig:
Just a question: How did you get the metanetx database? Is there some download somewhere/or some scrapping script which queries the whole database?

Midnighter:
@matthiaskoenig: http://www.metanetx.org/mnxdoc/mnxref.html

phantomas1234:
Closing this in favor of integrating solution into old cameo

PR #158, opened 2017-04-07
Feat/bring your own model old cameo

phantomas1234:
* Allow users to use their custom models and mappings
* Use metanetx_universal_model_bigg as default universal model
* Auto pick the compartment to connect to the universal model (if not provided)


phantomas1234:
@joaocardoso still getting cases where `max_predictions` is not respected but I think this should still be  merged.

PR #160, opened 2017-04-11
Feat/click cli

phantomas1234:
A cameo CLI using http://click.pocoo.org/5/


phantomas1234:
I will continue working on this after easter.

cli/cli.py
```diff
@@ -0,0 +1,181 @@
...
+    """cameo command line interface"""
+    pass
+
+
+@main.command()
+@click.option('-o', '--output', default='-', multiple=True, type=click.Path(),
+              help='Output filename. Multiple output files can be provided (pair with respective format options).')
+@click.option('-f', '--format', multiple=True, default=['csv'],
+              type=click.Choice(VALID_OUTPUT_FORMATS), help='Output file format (default csv).')
+@click.option('-h', '--host', default=['ecoli'], multiple=True,
```
joaocardoso: Default in the documentation doesn't match the default value. Can I pass multiple --host argument? I cannot figure it out from click documentation.

cameo/api/products.py
```diff
@@ -70,7 +73,15 @@ def search(self, query):
         if len(matches) > 0:
             return matches
         else:
-            raise Exception("No compound matches found for query %s" % query)
```
joaocardoso: Why shouldn't it raise an Exception (not necessarily `Exception` but a specific one) saying that it didn't found anything?

cameo/api/designer.py
```diff
@@ -197,28 +205,31 @@ def optimize_strains(self, pathways, view, aerobic=True):
...
+                      pathway.needs_optimization(model, objective=model.biomass)]
 
         print("Optimizing %i pathways" % len(strategies))
 
         results = DataFrame(columns=["host", "model", "manipulations", "heterologous_pathway",
                                      "fitness", "yield", "product", "biomass", "method"])
 
-        mapped_designs1 = view.map(opt_gene_runner, strategies)
-        mapped_designs2 = view.map(differential_fva_runner, strategies)
-        progress = ProgressBar(maxval=len(mapped_designs1) + len(mapped_designs2),
```
joaocardoso: How does the user keeps track of the progress?

cameo/api/designer.py
```diff
@@ -244,24 +251,23 @@ def build_results_data(self, strategy_designs, strategies, results, progress, of
...
-            _results['host'] = strategy[0].name
-            _results['model'] = strategy[1].id
+                _results.loc[j, 'heterologous_pathway'] = strategy['pathway']
+            _results['host'] = strategy['host']
+            _results['model'] = strategy['model'].id
             _results['fitness'] = fitness
             _results['yield'] = yields
             _results['biomass'] = biomass
             _results['product'] = target_flux
             _results['method'] = "DifferentialFVA+PathwayFinder"
```
joaocardoso: What if is a design from OptGene? This could be in the parameters.

cli/cli.py
```diff
@@ -0,0 +1,181 @@
...
+           max_pathway_predictions, differential_fva_points, pathway_prediction_timeout,
+           heuristic_optimization_timeout, logging):
+    """Compute strain designs for desired product."""
+    from cameo.api.designer import design
+    from cameo.api.hosts import hosts
+    from cameo.parallel import MultiprocessingView, SequentialView
+
+    logger.setLevel(logging)
+    cameo_logger.setLevel(logging)
+    inspyred_logger.setLevel(logging)
```
joaocardoso: We should not change inspyred logger. Debugging a heuristic optimization library is for advanced users, not for the high-level users.

cli/cli.py
phantomas1234: Right below your comment it says "Multiple hosts can be specified by repeating --host HOST". Which default doesn't match?

cli/cli.py
phantomas1234: I don't find this particular harmful to the user. The inspyred logger is only logging on DEBUG anyways. But I'll remove it.

cameo/api/products.py
phantomas1234: Because not finding something is not an error. I prefer to return an empty result.

cameo/api/designer.py
phantomas1234: This is only for the processing of the results, which really doesn't take that long so I don't see the point of having a progress bar.

cameo/api/designer.py
phantomas1234: Yes, I forgot to fix this one.

phantomas1234: Would be helpful if you could point out what I need to replace "DifferentialFVA+PathwayFinder" with?

PR #163, opened 2017-05-15
Feature: decompose coupled reactions optimization results

joaocardoso:
When the nullspace is used to calculate coupled reactions in the model, only one reaction is taken as representative of the "pathway". This is required to make the search space smaller and the algorithms faster.

Decomposition of the reaction sets into all combinations of equivalent substitutions recovers the designs masked by reducing the search candidates.

The users can then choose which designs are easier/convenient to implement on their own.


joaocardoso:
Looks like travis is failing due to some pickling problem.
https://travis-ci.org/biosustain/cameo/jobs/232387832#L714

http://stackoverflow.com/questions/36888228/python-read-pickle-importerror-no-module-named-indexes-base

Which versions of pandas are people using? I didn't change anything in travis....

cameo/flux_analysis/structural.py
```diff
@@ -149,32 +148,56 @@ def find_blocked_reactions_nullspace(model, ns=None, tol=1e-10):
...
-    group_dict = {i: {i} for i in range(len(non_blocked_reactions))}
-    for i, n in enumerate(range(len(non_blocked_reactions), len(non_blocked_reactions) + len(good_link))):
-        a = good_link[i]
-        group_dict[n] = group_dict[a[0]] | group_dict[a[1]]
-        del group_dict[a[0]]
-        del group_dict[a[1]]
-
-    groups = [frozenset(non_blocked_reactions[list(v)]) for v in group_dict.values() if len(v) > 1]
+    for i, reaction_i in enumerate(non_blocked_reactions):
+        reaction_i = non_blocked_reactions[i]
```
KristianJensen: This is a duplicate

cameo/flux_analysis/structural.py
```diff
@@ -149,32 +148,56 @@ def find_blocked_reactions_nullspace(model, ns=None, tol=1e-10):
...
+        left = non_blocked_ns[i]
+        group = next((g for g in groups if reaction_i in g), None)
```
KristianJensen: Won't this fail on python2.7? Use six.next

cameo/util.py
```diff
@@ -602,6 +602,32 @@ def reduce_reaction_set(reaction_set, groups):
...
+    list
+        A list of all possible group substitutions.
+
+    """
+    to_keep = []
+    to_replace = {}
+    for element in reactions:
+        for g in reaction_groups:
+            if element in g:
+                to_replace[element] = g.keys()
```
KristianJensen: There should probably be a break here. The if below could then be changed to an 'else' on the for loop

cameo/util.py
```diff
@@ -602,6 +602,32 @@ def reduce_reaction_set(reaction_set, groups):
...
+        if element not in to_replace:
+            to_keep.append(element)
+
+    return (list(combo) + to_keep for combo in itertools.product(*to_replace.values()))
```
KristianJensen: Nice :)

hredestig:
how about sending this directly to the devel06 branch?

hredestig:
Alright, unfortunate but I can understand that

phantomas1234:
@joaocardoso I will open a pull request with the branch that I have that eliminates all pickles in the codebase. Then we can finish this one here.

joaocardoso:
Ya, @KristianJensen already review the code. What is breaking now is pandas and pickled `DataFrame`s.

joaocardoso:
@phantomas1234 this one is good to go in too.

PR #167, opened 2017-05-16
Fix/upgrading services

hredestig:
fixes for upgrading model and iloop-to-model services
- use 0.18.1 pandas to temporarily resolve conflicts with saved pickles
- don't use unused argument `delete` in `remove_reactions`
- fixes for testing minho modeldb


hredestig:
codecov so noisy....

PR #168, opened 2017-05-17
Refactor/eliminate pickles

phantomas1234:


joaocardoso:
@phantomas1234 everything looks OK here too. This one is needed to fix #163.

Midnighter:
Should we merge this anyway? Seems like codecov got stuck somewhere.

PR #169, opened 2017-05-17
Refactor pandas.Panel result to multi-index dataframe in DifferentialFVA

phantomas1234:


Midnighter:
Locally all tests pass. When merging this we can unpin the pandas version again. I had to set it to 0.18.1 such that the pickle files can be loaded for testing. Ready for review IMHO @phantomas1234 but let's see what Travis has to say.

.travis.yml
```diff
@@ -38,7 +38,7 @@ before_install:
 - 'echo "this is a build for: $TRAVIS_BRANCH"'
 - 'if [[ "$TRAVIS_BRANCH" != "devel" ]]; then bash ./.travis/install_cplex.sh; fi'
 install:
-- pip install flake8 numpy scipy pyzmq pandas pytest pytest-cov
```
hredestig: why does numpy and pandas need to mentioned here, it's requirement in setup.py

Midnighter: I don't think it's necessary but that line was there before. I only pinned the pandas version.

requirements_dev.txt
```diff
@@ -8,7 +8,7 @@ tabulate
 numpy>=1.9.1
 scipy>=0.9.0
 blessings>=1.5.1
-pandas>=0.15.2
+pandas==0.18.1
```
joaocardoso: Are we fixing pandas? Or we should say >=0.18.1 (or actually latest version is 0.21.?)

phantomas1234: I think this in only in there for tests to pass because this branch still contains the pandas pickles. Should be set to `pandas>=0.21.1` eventually

joaocardoso: ok. But let's merge this in when its ready, then #168. We can update the version in #168.

joaocardoso:
@phantomas1234 can I merge this one? Tests are passing and I reviewed the code already.

PR #172, opened 2017-06-27
Fix/phenopp

hredestig:
- avoid mutables in defaults
- allow not specifying c-source (and failing to guess)


hredestig:
can't come up with a sensible test, only used when can't figure out the carbon source for the model for whatever reason

cameo/flux_analysis/analysis.py
```diff
@@ -506,7 +506,10 @@ def _get_c_source_reaction(model):
     source_reactions = [(reaction, reaction.flux * n_carbon(reaction)) for reaction in model_reactions if
```
Midnighter: Multiply by the stoichiometric coefficient as well? It **should** be 1 but you never know what people do with their models :wink: Or is that already part of `n_carbon`?

cameo/flux_analysis/analysis.py
```diff
@@ -506,7 +506,10 @@ def _get_c_source_reaction(model):
...
                         reaction.flux < 0]
     sorted_sources = sorted(source_reactions, key=lambda reaction_tuple: reaction_tuple[1])
```
Midnighter: `key=itemgetter(1)`

cameo/flux_analysis/analysis.py
hredestig: indeed, this mistake was actually original reason for the pr, fixed now

cameo/flux_analysis/analysis.py
hredestig: fixed, cheers

cameo/flux_analysis/analysis.py
```diff
@@ -499,14 +500,62 @@ def _get_c_source_reaction(model):
...
         return None
-    model_reactions = model.reactions.get_by_any(list(model.medium))
-    source_reactions = [(reaction, reaction.flux * n_carbon(reaction)) for reaction in model_reactions if
-                        reaction.flux < 0]
-    sorted_sources = sorted(source_reactions, key=lambda reaction_tuple: reaction_tuple[1])
-    return sorted_sources[0][0]
+    medium_reactions = model.reactions.get_by_any(list(model.medium))
+    medium_reactions_fluxes = [(rxn, total_carbon_flux(rxn, consumption=True)) for rxn in medium_reactions]
+    source_reactions = [(rxn, c_flux) for rxn, c_flux in medium_reactions_fluxes if c_flux > 0]
+    sorted_sources = sorted(source_reactions, key=itemgetter(1))
```
Midnighter: If only the maximum element is of interest, rather than sorting the container, I'd use `max(source_reactions, key=itemgetter(1))`. Didn't think about that before, sorry.

cameo/flux_analysis/analysis.py
```diff
@@ -318,7 +319,7 @@ def phenotypic_phase_plane(model, variables=[], objective=None, source=None, poi
         (defaults to the current model objective).
     source: Reaction or reaction identifier
         The reaction to use as the source when calculating mass and carbon yield. Set to the medium reaction with the
-        highest input carbon flux if left none.
+        highest input carbon flux if left None and
```
Midnighter: Seems unfinished...

cameo/flux_analysis/analysis.py
```diff
@@ -499,14 +500,62 @@ def _get_c_source_reaction(model):
...
+        flux for consumed metabolite, else produced
+
+    Returns
+    -------
+    tuple
+        metabolite, flux for the metabolite"""
+    if len(list(reaction.metabolites)) != 1:
+        raise ValueError('product flux only defined for single metabolite reactions')
+    met = list(reaction.metabolites)[0]
+    coeff = reaction.metabolites[met]
```
Midnighter:
```python
from six import iteritems
met, coeff = next(iteritems(reaction.metabolites))
```

cameo/flux_analysis/analysis.py
```diff
@@ -674,24 +705,15 @@ def mass_yield(self):
...
-        source_metabolite = list(self.source.metabolites)[0]
-        product_metabolite = list(self.product_reaction.metabolites)[0]
-        product_mass = product_metabolite.formula_weight
-        source_mass = source_metabolite.formula_weight
-        try:
-            mol_prod_mol_src = product_flux / (source_flux * -1)
-            return (mol_prod_mol_src * product_mass) / source_mass
-        except ZeroDivisionError:
-            return 0
+        mol_prod_mol_src = product_flux / source_flux
```
Midnighter: Can't remember if that's default in Python 2.7 now but to be safe you might want to add `from __future__ import division` to the top otherwise Python 2 will do integer division.

hredestig: Yup, thanks, fixed

cameo/flux_analysis/analysis.py
hredestig: Of course :) thanks

cameo/flux_analysis/analysis.py
hredestig: True, fixed, cheers

cameo/flux_analysis/analysis.py
hredestig: Of course, fixed

PR #175, opened 2017-07-20
Fix/cell fac course

hredestig:
Tons of changes here, best to click on each commit and view them separately as purpose is indicated that way (saves me making 10+ PRs..)

- sort various minor bugs (divide by zero and lingering usages of add_demand )
- move duplicated features with cobrapy out of cameo (pfba, find_esssential_x)
- remove `cameo.stuff`, will it be missed?


cameo/strain_design/heuristic/evolutionary/objective_functions.py
```diff
@@ -208,15 +208,13 @@ def __call__(self, model, solution, targets):
...
             product_flux = round(solution.fluxes[self.product], config.ndecimals)
             substrate_flux = round(sum(abs(solution.fluxes[s]) for s in self.substrates), config.ndecimals)
-        try:
+        if substrate_flux > config.non_zero_flux_threshold:
             return (biomass_flux * product_flux) / substrate_flux
-
-        except ZeroDivisionError:
-            return 0.0
+        else:
+            return 0.
```
joaocardoso: For now let's leave it like this, but it should return `self.worst_fitness` later. But for that we need to make the evaluation function aware of the direction of the optimization.

cameo/strain_design/deterministic/linear_programming.py
```diff
@@ -130,11 +131,11 @@ def _reduce_to_nullspace(self, reactions):
     def _build_problem(self, essential_reactions, use_nullspace_simplification):
         logger.debug("Starting to formulate OptKnock problem")
 
-        self.essential_reactions = find_essential_reactions(self._model) + self._model.exchanges
+        self.essential_reactions = find_essential_reactions(self._model).union(self._model.exchanges)
```
joaocardoso: Does the model still has the `essential_reactions`? If yes, then it should call `self._model.essential_reactions()`

cameo/strain_design/heuristic/evolutionary/objective_functions.py
hredestig: Alright, this PR was about getting rid of the ZeroDivError so yep, let's handle the improvement in a different pr

cameo/strain_design/deterministic/linear_programming.py
hredestig: No, `essential_reactions` is not a method of cobra `Model` didn't really see the need for it there..

hredestig:
Sure thing, plotting_old is removed

PR #177, opened 2017-07-24
fix: 0/0 shoud be equivalent to 1 if the ratio of non infinite values…

joaocardoso:
… is 1.

Example:
```
a = [1, 2, 4, 0, 1, 5, 6, 7]
b = a

a/b == 1, but 0/0 == inf
```

### DIFF
--- a/cameo/flux_analysis/structural.py
+++ b/cameo/flux_analysis/structural.py
@@ -191,6 +191,25 @@ def find_coupled_reactions_nullspace(model, ns=None, tol=1e-10):
             reaction_j = non_blocked_reactions[j]
             right = non_blocked_ns[j]
             ratio = np.apply_along_axis(lambda x: x[0] / x[1] if abs(x[1]) > 0. else np.inf, 0, np.array([left, right]))
+
+            # special case:
+            # if ratio is 1 (a/b == 1) then a == b.
+            # but if a = 0 and b = 0, then a/b = np.inf.
+            # solution:
+            # mask inf from ratio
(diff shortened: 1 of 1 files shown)


cameo/flux_analysis/structural.py
```diff
@@ -191,6 +190,13 @@ def find_coupled_reactions_nullspace(model, ns=None, tol=1e-10):
             reaction_j = non_blocked_reactions[j]
             right = non_blocked_ns[j]
             ratio = np.apply_along_axis(lambda x: x[0] / x[1] if abs(x[1]) > 0. else np.inf, 0, np.array([left, right]))
+
```
hredestig: I find this all pretty cryptic, would do something like 
```
def arr_div(left, right):
    def inf_div(x, tol=1e-6):
        if not np.isfinite(x[0]) and not np.isfinite(x[0]):
            return 1
        elif not np.isfinite(x[0]) or not np.isfinite(x[0]):
            return 0
        elif abs(x[0]) < tol and abs(x[0]) < tol:
            return 1
        elif abs(x[1]) < tol:
            return 0
        else:
            return x[0] / x[1]
    return np.apply_along_axis(inf_div, 0, np.array([left, right]))
ratio = arr_div(left, right, tol * 100) 
```

joaocardoso: It cannot be done in the same step, because it only applies if the ratio between all other elements is 1 and it is only true if `left[i] == 0` and `right[i] == 0`.

PR #207, opened 2018-04-24
Fix issue with pip 10.0

KristianJensen:
Pip does not support use as a library. https://github.com/pypa/pip/issues/5243

### DIFF
--- a/cameo/util.py
+++ b/cameo/util.py
@@ -30,7 +30,7 @@ from uuid import uuid1
 
 import numpy
 import pandas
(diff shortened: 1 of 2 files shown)


Midnighter:
Since we use a similar setup in many packages I'd like to extract the basic functionality into a tiny package. Better to quickly fix this for now though so thank you for doing that :smiley:

Midnighter:
Changes are part of #208 so I'm closing this one.

PR #208, opened 2018-05-11
fix: repair Travis builds

Midnighter:
* reduce dependencies
* drop Python 3.4
* make use of new CPLEX


Midnighter:
I had to include all changes from #207 in order to make this work.

Midnighter:
Just a note: I set up the installation of CPLEX only for the master branch and tagged commits so I couldn't test that yet. I *hope* it will work, though.

PR #215, opened 2018-09-24
GrowthCouplingPotential

KristianJensen:
aka OptCouple
This PR adds my growth coupling potential based MILP optimization to the strain_design.deterministic.linear_programming module
I’ve also added some tests, but I don’t really know how pytest works, so I’ve just tried to mirror the OptKnock tests


tests/test_strain_design_deterministic.py
```diff
@@ -146,3 +160,31 @@ def test_invalid_input(self, cplex_optknock):
...
+        assert "obj_val" in result
+        assert "knockouts" in result
+        assert "knockins" in result
+        assert "medium" in result
+
+    def test_growth_coupling_potential_benchmark(self, gurobi_growth_coupling_potential, benchmark):
+        _, growth_coupling_potential = gurobi_growth_coupling_potential
+        benchmark(growth_coupling_potential.run)
+
+    def test_result_is_correct(self, gurobi_growth_coupling_potential):
```
Midnighter: As far as I can tell, this test does not verify the function of the method. TPI is already essential for growth and the knock-out results do not modify that situation. (Please see the attached notebook.)
[opt_couple.ipynb.gz](https://github.com/biosustain/cameo/files/3378146/opt_couple.ipynb.gz)

cameo/strain_design/deterministic/linear_programming.py
```diff
@@ -426,3 +430,394 @@ def _repr_html_(self):
...
+        super(GrowthCouplingPotential, self).__init__(**kwargs)
+
+        # Check model for forced fluxes, which should not be present. The zero solution must be allowed.
+        for r in model.reactions:
+            if r.lower_bound > 0 or r.upper_bound < 0:
+                raise ValueError(
+                    "%s has a forced flux. Remove the reaction or change its bounds for the function to work." % r.id)
+
+        # If a cobra.Reaction is given as target, convert to its ID string.
+        if isinstance(target, cobra.core.Reaction):
```
Midnighter:
```suggestion
        if isinstance(target, cobra.Reaction):
```

cameo/strain_design/deterministic/linear_programming.py
```diff
@@ -426,3 +430,394 @@ def _repr_html_(self):
...
+            target = target.id
+        self.target = target
+
+        if target not in knockout_reactions:
+            knockout_reactions.append(target)
+
+        self.original_model = model
+        self.biomass_id = [r for r in model.reactions if r.objective_coefficient != 0][0].id
```
Midnighter: `objective_coefficient` is unfortunately deprecated. Maybe this could be inferred from the objective expression or specifically required as a parameter.

cameo/strain_design/deterministic/linear_programming.py
```diff
@@ -426,3 +430,394 @@ def _repr_html_(self):
...
+        else:
+            logger.warning("You are trying to run 'GrowthCouplingPotential' with %s. This might not end well." %
+                           model.solver.interface.__name__.split(".")[-1])
+
+        # Remove reactions that are blocked: no flux through these reactions possible.
+        # This will reduce the search space for the solver, if not done already.
+        if remove_blocked:
+            blocked_reactions = cameo.flux_analysis.analysis.find_blocked_reactions(model)
+            model.remove_reactions(blocked_reactions)
+            blocked_reaction_ids = [r.id for r in blocked_reactions]
```
Midnighter: Turning this into a set will improve the following statements.

```suggestion
            blocked_reaction_ids = {r.id for r in blocked_reactions}
```

cameo/strain_design/deterministic/linear_programming.py
```diff
@@ -426,3 +430,394 @@ def _repr_html_(self):
...
+            knockout_reactions = [r for r in knockout_reactions if r not in blocked_reaction_ids]
+            knockin_reactions = [r for r in knockin_reactions if r not in blocked_reaction_ids]
+            medium_additions = [r for r in medium_additions if r not in blocked_reaction_ids]
+            logger.debug("Removed " + str(len(blocked_reactions)) + " reactions that were blocked")
+
+        # Make dual
+        copied_model = model.copy()
```
Midnighter: Are the solver tolerances that were set above actually copied over?

cameo/strain_design/deterministic/linear_programming.py
```diff
@@ -426,3 +430,394 @@ def _repr_html_(self):
...
+        if isinstance(target, cobra.core.Reaction):
+            target = target.id
+        self.target = target
+
+        if target not in knockout_reactions:
+            knockout_reactions.append(target)
+
+        self.original_model = model
+        self.biomass_id = [r for r in model.reactions if r.objective_coefficient != 0][0].id
+        self.model = model = model.copy()
```
Midnighter: Assigning to `self.model` seems to have no effect since another model copy is assigned on line 691.
```suggestion
        model = model.copy()
```

cameo/strain_design/deterministic/linear_programming.py
```diff
@@ -426,3 +430,394 @@ def _repr_html_(self):
...
+                model.solver = "gurobi"
+            # The tolerances are set to the minimum value. This gives maximum precision.
+            problem = model.solver.problem
+            problem.params.NodeMethod = 1  # primal simplex node relaxation
+            problem.params.FeasibilityTol = 1e-9
+            problem.params.OptimalityTol = 1e-3
+            problem.params.IntFeasTol = 1e-9
+            problem.params.MIPgapAbs = 1e-9
+            problem.params.MIPgap = 1e-9
+            # problem.params.TimeLimit = 200 # Use max 200 seconds when called, return best solution after that
```
Midnighter: What happened to these parameters? If they are not important, can these lines be removed?

cameo/strain_design/deterministic/linear_programming.py
```diff
@@ -426,3 +430,394 @@ def _repr_html_(self):
...
+            problem.parameters.mip.strategy.startalgorithm.set(1)  # primal simplex node relaxation
+            problem.parameters.simplex.tolerances.feasibility.set(
+                1e-9)
+            problem.parameters.simplex.tolerances.optimality.set(
+                1e-3)
+            problem.parameters.mip.tolerances.integrality.set(
+                1e-9)
+            problem.parameters.mip.tolerances.absmipgap.set(1e-9)
+            problem.parameters.mip.tolerances.mipgap.set(1e-9)
+            # problem.parameters.mip.pool.relgap.set(0.9)
```
Midnighter: Same question here.

cameo/strain_design/deterministic/linear_programming.py
```diff
@@ -426,3 +430,394 @@ def _repr_html_(self):
...
+        native_y_vars = {}  # 1 for knockout, 0 for active
+        heterologous_y_vars = {}  # 1 for knockin, 0 for inactive
+        medium_y_vars = {}  # 1 for medium addition (up to -10), 0 for no addition
+
+        # Now the fun stuff
+        constrained_dual_vars = set()
+
+        # Fill the dictionaries with binary variables
+        # For the knockouts: only for the native reactions which are not exchanges
+        # for reaction in [r for r in model.reactions - excluded_reactions if r.type == "native"]:
```
Midnighter:
```suggestion
```

cameo/strain_design/deterministic/linear_programming.py
```diff
@@ -426,3 +430,394 @@ def _repr_html_(self):
...
+        for reac_id in knockout_reactions:
+            reaction = model.reactions.get_by_id(reac_id)
+
+            # Add constraint variables
+            interface = model.solver.interface
+            y_var = interface.Variable("y_" + reaction.id, type="binary")
+
+            # Constrain the primal: flux through reactions within (-1000, 1000)
```
Midnighter: This might depend on `cobra.Configuration.bounds`?

cameo/strain_design/deterministic/linear_programming.py
```diff
@@ -426,3 +430,394 @@ def _repr_html_(self):
...
+            native_y_vars[y_var] = reaction
+
+        # For the knockins and medium additions:
+        for reac_id in knockin_reactions + medium_addition_reactions:
+            reaction = model.reactions.get_by_id(reac_id)
+            # Add constraint variables
+            interface = model.solver.interface
+            y_var = interface.Variable("y_" + reaction.id, type="binary")
+
+            # Constrain the primal: flux through reaction within (-1000, 1000)
```
Midnighter: Same question as above.

cameo/strain_design/deterministic/linear_programming.py
```diff
@@ -426,3 +430,394 @@ def _repr_html_(self):
...
+
+        self.use_solution_pool = use_solution_pool
+
+        self.model.solver.configuration.presolve = True
+
+        if use_solution_pool:
+            if "gurobi_interface" not in model.solver.interface.__name__:
+                raise NotImplementedError("Solution pools are currently only available with the Gurobi solver")
+
+            self.model.solver.problem.params.PoolSearchMode = 2
```
Midnighter: A comment here on what this means would be nice.
https://www.gurobi.com/documentation/8.1/refman/poolsearchmode.html

tests/test_strain_design_deterministic.py
```diff
@@ -146,3 +160,31 @@ def test_invalid_input(self, cplex_optknock):
...
         with pytest.raises(ValueError):
             optknock.run(biomass="Biomass_Ecoli_core_N_lp_w_fsh_GAM_rp__Nmet2")
+
+
+@pytest.mark.skipif('gurobi' not in solvers, reason="No gurobi interface available")
+class TestGrowthCouplingPotential:
+    def test_growth_coupling_potential_runs(self, gurobi_growth_coupling_potential):
+        _, growth_coupling_potential = gurobi_growth_coupling_potential
+        result = growth_coupling_potential.run()
+        assert "obj_val" in result
```
Midnighter: This seems like it's not a very meaningful test. As long as there is no exception in the `run` method, these keys will always be present.

cameo/strain_design/deterministic/linear_programming.py
```diff
@@ -426,3 +430,394 @@ def _repr_html_(self):
...
+    medium_additions : list
+        A list of metabolite ID's for the the metabolites that should potentially be added to the medium
+    n_knockouts : int
+        The maximum allowed number of knockouts
+    n_knockins : int
+        The maximum allowed number of knockins
+    n_medium : int
+        The maximum allowed number of medium additions
+    use_solution_pool : Bool
+        Set to True when using the Gurobi solver to identify multiple alternative solutions at once
```
Midnighter: A reference to the published paper in the docstring would be very useful.

cameo/strain_design/deterministic/linear_programming.py
```diff
@@ -426,3 +430,394 @@ def _repr_html_(self):
...
+    Optimize the Growth Coupling Potential through a combination of knockouts, knockins and medium additions
+
+    Use the 'run' method to optimize the MILP problem and find the best solution.
+
+    In order to find multiple sub-optimal solutions through the use of solution pools, please refer to the
+    relevant solvers documentation. The underlying solver object can be accessed through
+        GrowthCouplingPotential(...).model.solver.problem
+
+
+    Parameters
```
Midnighter:
```suggestion
    Attributes
```

bdelepine:
Hello!

Do you have any update on this PR? Is it still alive? Is it safe to use this branch to try out OptCouple?

KristianJensen:
Hi @bdelepine 

The PR is a bit stale as I haven't had time to finalize the unittests and incorporate the suggested changes. But the method should work. If you encounter any problems, feel free to post them here.

cameo/strain_design/deterministic/linear_programming.py
```diff
@@ -426,3 +430,394 @@ def _repr_html_(self):
...
+    ----------
+    model : cobra.Model
+        A cobra model object containing the target reaction as well as all native and heterologous reactions.
+    target : cobra.Reaction or string
```
bdelepine: This differ from the signature of OptKnock class and its run method: 
```
OptKnock(model, exclude_reactions=None, remove_blocked=True, fraction_of_optimum=0.1, exclude_non_gene_reactions=True, use_nullspace_simplification=True, *args, **kwargs)
run(self, max_knockouts=5, biomass=None, target=None, max_results=1, *args, **kwargs)
```
An OptKnock object is created without mention of the target, but it is a parameter set in the `run()` method. Same remark for the number of knockouts. I suggest that it should be harmonized one way or the other in both classes, I guess keeping OptKnock strategy for backward compatibility.

bdelepine:
Hi @KristianJensen,

Sadly, I confirm that the current branch does not work. The `sol.f` triggered an AttributeError. Once corrected, the code runs (using CPLEX). Unfortunately, I was still unable to get results since I keep getting `{'knockouts': [], 'knockins': [], 'medium': [], 'obj_val': 0.0}` despite trying different set of parameters. I guess my example is broken or that something is silently off somewhere 😕.

HTH, +

cameo/strain_design/deterministic/linear_programming.py
KristianJensen: I'm not sure, but that does not matter since the dual model is only constructed in order to copy stuff and add it back to the main model.

cameo/strain_design/deterministic/linear_programming.py
KristianJensen: Thanks for pointing that out. Regarding the number of knockouts etc it should be relatively easy to change to follow the OptKnock signature. The target is a bit more tricky. The reason I put it as an argument to `__init__` is that the choice of target reaction affects the construction of the optimization problem more than is the case for OptKnock. I will think about whether/how this can be changed

KristianJensen:
Thanks for the report @bdelepine 
I will try to get this PR finished ASAP - I will also need to use it myself in my current work.
I have now addressed some of @Midnighter's comments and updated the code to be compatible with modern cobrapy. I will do some testing of the method next.

If you want to try the method in the meantime, there is an implementation in the supplementary material of the paper, including a list of package versions.

cameo/strain_design/deterministic/linear_programming.py
bdelepine: Honestly, it might be better for cameo to accept that the signature between optimisation classes (and there associated run method)  is **expected** to be different for each class. This would give more flexibility to create the best preprocessing environment for the problem, just like you did. However this is far from an "object-oriented" mindset, which can be surprising for users. So, an acceptable solution IMHO would be to simply warn users in cameo documentation, but really, this is a design decision that should be made by package maintainers since the question will arise for each new algorithm implemented in cameo.

cameo/strain_design/deterministic/linear_programming.py
KristianJensen: Yes, this should be taken into account now

KristianJensen:
@Midnighter I have pushed some changes:
- The code is now compatible with modern cobrapy
- The test no longer relies on gurobi, but runs with glpk
- I've changed the test to a target that is not already growth-coupled (and added an assertion to verify this)

Let me know if there are other issues I should address (or if I should ping someone else instead :) )

carrascomj (approved):
Pinging requester @KristianJensen 
I merged latest changes into this branch, CI is passing and I think is ready for merging into devel.

phantomas1234:
Thanks @carrascomj!

PR #218, opened 2018-12-05
 Update simulation.py to use solution.objective_value vs. legacy solution.f

jlerman44:
Update simulation.py to use solution.objective_value vs. legacy solution.f

Tested lightly

### DIFF
--- a/cameo/flux_analysis/simulation.py
+++ b/cameo/flux_analysis/simulation.py
@@ -178,7 +178,8 @@ def moma(model, reference=None, cache=None, reactions=None, *args, **kwargs):
         solution = model.optimize(raise_error=True)
 
         if reactions is not None:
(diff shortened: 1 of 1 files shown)


phantomas1234:
Hi @jlerman44, thanks for contributing! Could you run `flake8 .` (`pip install flake8` if you don't have it yet) in the cameo directory of your fork? Maybe it is just an overlong line?

jlerman44:
Oh yes. Let me do that! Sorry, i put in a quick fix using github's code editor. Did not have my linter on!

phantomas1234:
I'll check the broken build tomorrow, I don't quite remember if we ever got around making travis-ci builds work with forks. The problem was that we're installing cplex in the build which we can only do for the main repo.

PR #228, opened 2019-02-04
Update path to pandas ipynb test

kvikshaug:
In pandas 0.24.0 and above, the internal `in_ipython_frontend` function
has been moved. This updates the reference, but it is still an internal
module and may break at any time so consider replacing it with something
more permanent.


phantomas1234:
Merging this because it is just flake8 (which is going to be fixed in #229)

PR #234, opened 2019-02-20
fix: predicted pathways containing exch reactions for native compounds

### LINKED ISSUE #233: Boundary reactions in predicted pathway results can lead to higher growth rate
A pathway result defines demand reactions for all pathway metabolites in `.exchanges` and additionally defines more demand reactions as knock-in targets in `.targets`.

Adding a demand reaction on protons combined with typically unlimited proton uptake rates can yield unrealistic results (higher than wildtype growth) due to proton pumping.

Suggested changes:

* Do not define any demand reactions for native metabolites (those for which adapter reactions exist `.adapters`).
* Do not include demand reactions as knock-in targets.

phantomas1234:
Fixes #233

A change of `model.exchanges` in upstream cobrapy has led to the
incorporation of DM_ reactions for native compounds (H+, H2O, CO2, etc.)
into the `PathwayPredictor.model` (host model + universal model). The
reason for this was that previously `predictor.universal_model.exchanges`
returned a list of all DM_... reactions included but after the change in cobrapy
it returned `[]`. This list is used to remove DM_ reactions for native compounds.

I have replaced `model.exchanges` with `model.boundary` (which has
the same behavior as `model.exchanges` had in the past) to fix this problem.

### DIFF
--- a/cameo/flux_analysis/analysis.py
+++ b/cameo/flux_analysis/analysis.py
@@ -169,7 +169,7 @@ def find_blocked_reactions(model):
 
     """
     with model:
-        for exchange in model.exchanges:
+        for exchange in model.boundary:
             exchange.bounds = (-9999, 9999)
         fva_solution = flux_variability_analysis(model)
     return frozenset(
--- a/cameo/flux_analysis/util.py
+++ b/cameo/flux_analysis/util.py
@@ -49,7 +49,7 @@ def remove_infeasible_cycles(model, fluxes, fix=()):
     """
     with model:
         # make sure the original object is restored
-        exchange_reactions = model.exchanges
+        exchange_reactions = model.boundary
         exchange_ids = [exchange.id for exchange in exchange_reactions]
         internal_reactions = [reaction for reaction in model.reactions if reaction.id not in exchange_ids]
         for exchange in exchange_reactions:
--- a/cameo/strain_design/deterministic/flux_variability_based.py
+++ b/cameo/strain_design/deterministic/flux_variability_based.py
@@ -197,8 +197,8 @@ class DifferentialFVA(StrainDesignMethod):
         reference_blocked_reactions = find_blocked_reactions_nullspace(self.reference_model, self.reference_nullspace)
         self.exclude += [reaction.id for reaction in reference_blocked_reactions]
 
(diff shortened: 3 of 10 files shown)


cameo/flux_analysis/analysis.py
```diff
@@ -169,7 +169,7 @@ def find_blocked_reactions(model):
 
     """
     with model:
-        for exchange in model.exchanges:
+        for exchange in model.boundary:
```
Midnighter: I would remove this entire function if easily possible. It is identical to the cobrapy version (which uses `model.exchanges`).

cameo/flux_analysis/util.py
```diff
@@ -49,7 +49,7 @@ def remove_infeasible_cycles(model, fluxes, fix=()):
     """
     with model:
         # make sure the original object is restored
-        exchange_reactions = model.exchanges
+        exchange_reactions = model.boundary
```
Midnighter: `exchange_ids` not being a set makes me :sob: :panda_face:

cameo/strain_design/deterministic/flux_variability_based.py
```diff
@@ -856,7 +856,7 @@ def run(self, target=None, max_enforced_flux=0.9, number_of_results=10, exclude=
         ndecimals = config.ndecimals
 
         # Exclude list
-        exclude = list(exclude) + model.exchanges
+        exclude = list(exclude) + model.boundary
```
Midnighter: `exclude_ids` would benefit from being turned into a set after line 865.

cameo/strain_design/pathway_prediction/pathway_predictor.py
```diff
@@ -421,7 +421,7 @@ def _add_switches(self, reactions):
         self._y_vars_ids = [var.name for var in y_vars]
 
     def _extend_model(self, original_exchanges):
-        for exchange in self.model.exchanges:
+        for exchange in self.model.boundary:
```
Midnighter: Might be `exchanges` here. Not sure how it interacts with the universal model.

phantomas1234:
@Midnighter I just fixed the broken use of `model.exchanges`. Nothing more. Obviously, the whole medium loading functionality doesn't make sense anymore in cameo, so feel free to improve in a future pull request.

PR #240, opened 2019-06-19
Fix issues in differential FVA

Midnighter:
A number of issues surfaced with differential FVA:

1. Correct normalization of flux ranges
2. Correct the conditions for suddenly essential reactions
3. Exclude cases where the envelope is computed based on more than one variable

Please note that two test cases fail because now the reference files are, of course, incorrect. @phantomas1234 were those references generated with Mathematica?


phantomas1234:
@Midnighter  I don't think the test data came from the Mathematica implementation. Can you run the PR version of DifferentialFVA on the serine production case and then we compare with the Mathematica notebook to be sure it is correct now?

phantomas1234:
Also why is there no travis-ci build on this?

Midnighter:
The serine case is running. We can look at results tomorrow.

> Also why is there no travis-ci build on this?

Since this PR is targetting the other branch, I think the current Travis config disables it.

cameo/strain_design/deterministic/flux_variability_based.py
```diff
@@ -369,8 +440,8 @@ def run(self, surface_only=True, improvements_only=True, progress=True, view=Non
             ] = True
 
             df.loc[
-                ((df.lower_bound <= 0) & (df.lower_bound > 0)) | (
-                    (ref_lower_bound >= 0) & (df.upper_bound <= 0)),
+                (zero_overlap_mask & (df.lower_bound > 0)) | (
```
phantomas1234: How reliable is this in terms of numerical issues? Should this maybe use `non_zero_flux_threshold` or similar?

cameo/strain_design/deterministic/flux_variability_based.py
```diff
@@ -344,6 +410,11 @@ def run(self, surface_only=True, improvements_only=True, progress=True, view=Non
         ref_lower_bound = self.reference_flux_ranges.lower_bound.apply(
             lambda v: 0 if abs(v) < non_zero_flux_threshold else v)
```
phantomas1234: is `ref_lower_bound` obsolete now? Maybe it should be used for `zero_overlap_mask` below?

cameo/strain_design/deterministic/flux_variability_based.py
phantomas1234: I mean something like `zero_overlap_mask & (df.lower_bound > non_zero_flux_threshold)`

Midnighter: What about modifying all solutions as in the following:
```py
df[df[["lower_bound", "upper_bound"]].abs() < non_zero_flux_threshold] = 0
```

cameo/strain_design/deterministic/flux_variability_based.py
Midnighter: Right, I didn't take care of this.

cameo/strain_design/deterministic/flux_variability_based.py
phantomas1234: fine with me

Midnighter:
Alright, coverage decreases a bit but overall it's an improvement.

PR #248, opened 2019-10-03
Dropping Python2 support

carrascomj:
I've removed six related code so Python2 shouldn't be supported anymore.


carrascomj:
Ok, I am not very familiar with travis. It should be fine now.

PR #270, opened 2020-12-12
Replace Travis with github actions

carrascomj:
Use github actions for CI/CD instead of travis. This PR makes cameo's CI consistent with the cobrapy ecosystem and it avoids some issues with travis.

This PR tries to replicate the previous setup with travis. Further improvements should be explored in future PRs (black, isort, python/os versions...).


.github/workflows/main.yml
```diff
@@ -0,0 +1,48 @@
...
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        sudo apt install openbabel
+        python -m pip install --upgrade pip setuptools wheel
+        python -m pip install tox tox-gh-actions
```
carrascomj, line 42: Py3.6 `.travis/install_cplex.sh`:
- Mock and check that secret is not showed in log.
- Adapt it to python3.8.

carrascomj (commented):
Required changes before merging.

.github/workflows/main.yml
carrascomj: CPLEX path and unpacking must be corrected.

phantomas1234 (approved):
LGTM, thanks!

PR #272, opened 2021-03-01
Fix plotting

carrascomj:
Closes #271 
### Description
[Escher](https://github.com/zakandrewking/escher/) deprecated the `display_on_notebook` and `display_in_browser` functions in favor of using widgets.

### Implementation
* Updated all escher outputs; the methods `display_on_notebook` and `display_in_browser` has been replaced by display(builder) and builder.save_html(), respectively (as explained in the docstrings of [the deprecated functions](https://github.com/zakandrewking/escher/blob/master/py/escher/plots.py#L588)).
* Added documentation about jupyter widgets on the installation section (also for plotly).
* Fixed one plotting function on DIfferentialFVA.

PR #273, opened 2021-03-26
Refactor bounds behaviour

carrascomj:
Related to https://github.com/opencobra/cobrapy/issues/793
Merge this after #272 

### Description
Previously in cobrapy, setting a lower bound greater than the upper bound (or viceversa) set both bounds to the input value. This was removed in https://github.com/opencobra/cobrapy/pull/1058 and it will raise an error if the bounds are inconsistent.

### Implementation
* [`.bounds`](https://cobrapy.readthedocs.io/en/latest/autoapi/cobra/core/reaction/index.html#cobra.core.reaction.Reaction.bounds) was used wherever the intention is to fix the reaction flux to a value.
* Some tests that checked the old behavior were removed or changed.

### DIFF
--- a/tests/test_solver_based_model.py
+++ b/tests/test_solver_based_model.py
@@ -187,7 +187,8 @@ class TestReaction:
         assert acald_reaction.forward_variable.ub == 1000.
         assert acald_reaction.reverse_variable.lb == 0
         assert acald_reaction.reverse_variable.ub == 1000.
-        acald_reaction.bounds = (acald_reaction.lower_bound - 100,) * 2
+        fixed_flux = acald_reaction.lower_bound - 100
+        acald_reaction.bounds = (fixed_flux, fixed_flux)
         assert acald_reaction.lower_bound == -1100.0
         assert acald_reaction.upper_bound == -1100.0
         assert acald_reaction.forward_variable.lb == 0
@@ -208,8 +209,8 @@ class TestReaction:
         assert reac.lower_bound == 0
         assert reac.upper_bound == 0
         reac.bounds = (-2, -2)
-        assert reac.lower_bound == -2
-        assert reac.upper_bound == -2
+        assert reac.forward_variable.lb == 0
+        assert reac.forward_variable.ub == 0
         reac.bounds = (2, 2)
         assert reac.reverse_variable.lb == 0
         assert reac.reverse_variable.ub == 0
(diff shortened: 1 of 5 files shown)


Midnighter:
@carrascomj can I assist you in getting this merged? I'd like to make a new cobrapy release and this is blocking me. Alternatively, can we pin cobrapy to a specific version in cameo so that it doesn't break?

carrascomj:
@Midnighter Yes, please. This and the parent PR (#272, some docs and escher) are both ready to review. It would wait until cobrapy depends on the latest optlang to pin it here.

Midnighter:
Moving to latest optlang is exactly why I want to make a new cobrapy release 😆

tests/test_solver_based_model.py
```diff
@@ -187,7 +187,7 @@ def test_set_bounds_scenario_1(self, core_model):
         assert acald_reaction.forward_variable.ub == 1000.
         assert acald_reaction.reverse_variable.lb == 0
         assert acald_reaction.reverse_variable.ub == 1000.
-        acald_reaction.upper_bound = acald_reaction.lower_bound - 100
+        acald_reaction.bounds = (acald_reaction.lower_bound - 100,) * 2
```
Midnighter, line 190: I have the suspicion that writing out the full tuple instead of `* 2` is easier to grok. I know it creates a pair but it might be misunderstood since conceptually we're dealing with numerical values here.

tests/test_solver_based_model.py
```diff
@@ -256,7 +232,7 @@ def test_set_bounds_scenario_2(self, core_model):
         assert acald_reaction.forward_variable.ub == 1000.
         assert acald_reaction.reverse_variable.lb == 0
         assert acald_reaction.reverse_variable.ub == 1000.
-        acald_reaction.lower_bound = acald_reaction.upper_bound + 100
+        acald_reaction.bounds = (acald_reaction.upper_bound + 100,) * 2
```
Midnighter, line 235: Same comment about readability.

tests/test_solver_based_model.py
```diff
@@ -204,39 +204,15 @@ def test_set_bounds_scenario_1(self, core_model):
...
-        reac.lower_bound = 2
-        assert reac.lower_bound == 2
-        assert reac.upper_bound == 2
-        assert reac.forward_variable.lb == 2
-        assert reac.forward_variable.ub == 2
         reac.knock_out()
-        reac.upper_bound = -2
+        assert reac.lower_bound == 0
+        assert reac.upper_bound == 0
+        reac.bounds = (-2, -2)
```
Midnighter, line 210: Should the same be tested as below?

```py
        assert reac.variable.lb == 0
        assert reac.variable.ub == 0
```

Midnighter (approved):
Some minor comments, otherwise this looks good 🙂

PR #285, opened 2021-10-07
fix: set explicitly the auth tokens for twine

carrascomj:
### Description

Twine upload failed. It might be a) something obscure with the keyring/env variables not being read; or b) the authentication password actually changed.

### Implementation

Tackling a), explicitly set the password and user as command line arguments in the `twine upload` command.


phantomas1234:
Thanks @carrascomj !

PR #287, opened 2021-10-31
Move escher to extras

matthiaskoenig:
Escher has very old dependencis such as Jinja2 < 3 which is breaking many other packages. Moving this to extras to allow using cameo with other packages.


carrascomj:
I agree with the change but it requires implementing some try-except blocks (like in https://github.com/biosustain/cameo/blob/devel/cameo/flux_analysis/simulation.py#L486) because some modules import escher by default.

I believe that https://github.com/biosustain/cameo/blob/devel/cameo/visualization/visualization.py and https://github.com/biosustain/cameo/blob/devel/cameo/strain_design/deterministic/flux_variability_based.py#L822 (the latter uses an import form escher_ext) are the culprits of the top-level imports.

